Issue #24, opened 2016-12-22
Mine Switcher: Delete state when switching mines


yochannah:
Items which should have no content when mines are switched (but currently do)  
 
- Region Search
- Template Filters
- Upload
- Search

Most of these items make *no* sense when in the context of a different Mine

PR #62, opened 2017-01-19
deep links fixed, id resolver preview works again, removed distribution from regionsearch

yochannah:
... hopefully.

### DIFF
--- a/less/components/idresolver.less
+++ b/less/components/idresolver.less
@@ -192,7 +192,7 @@
                 .button-color;
             }
 
-            span {
+            .id-item {
                 padding-left: 5px;
             }
         }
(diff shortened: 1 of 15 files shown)


yochannah:
- deep links fixed #62 
- id resolver preview works again 
- removed distribution from regionsearch
- generified id resolver to accept types other than genes 
- no longer return 401 when requesting an anon token. @joshkh - notes here on how to test this one: https://github.com/intermine/redgenes/issues/21  
- don't break whole app if user navigates to /#results when there are no results in the db. #65
- remove stateful keys from db when we switch mines #24 
- Don't allow IDresolver "view all" button to be clicked until there are some good results  #63 
- Fix styling for report page to not shrink to content when it's really tiny or expand wider than the screen #64 
- resolving duplicate/ambiguous ids in the id resolver should work for proteins and hopefully other types too now

joshkh:
Great job! Spot checked and merged.

PR #105, opened 2017-03-10
Example queries for Query Builder

joshkh:
Add example queries for query builder, clear query builder settings when switching mines


yochannah:
Tested: Results tables work ok in flymine beta but not other mines

yochannah (approved):
There seems to be a problem with enrichment on flymine beta but not humanmine beta (??). It existed before this change so I'll merge and create a ticket.

Issue #88, opened 2017-02-27
List upload / id resolver: handle ids correctly when the selected organism is changed


yochannah:
from @julie-sullivan 

I uploaded a list of dmel but it was still set on mouse from me goofing around a bit. Can changing the organism refresh the page like changing the type does?

yochannah:
@julie-sullivan has highlighted that this is still a problem. If there is an ambiguous ID for, say, fly and human, then jumping between fly and human will work, but we aren't removing the ids that are unresolved - so if you enter a fly gene then change the organism to D. rerio, most of the ids should turn red but they don't.

PR #107, opened 2017-03-13
Update ID Resolver when Organism Changes

joshkh:
1. Update ID Resolver when Organism Changes
2. Add default query builder example for YeastMine


joshkh:
Fixes #88

yochannah:
This can't be merged until 88 is fixed-er :cry:

PR #110, opened 2017-03-14
Update ID Resolver when Organism Changes

joshkh:
This PR includes the Enrichment Query fix so you might want to merge it last. The relevant commit for updating the ID resolver when the organism changes is the last commit, "Clear previous results when organism value changes"


joshkh:
Fixes issue #88

yochannah (approved):

PR #112, opened 2017-03-14
Add human, fly, and mouse mines as they're 1.7.1

yochannah:
plus fix a regression due to a merge conflict that made all the QB examples disappear.


joshkh (commented):
Looks good!

Issue #21, opened 2016-12-15
Kill anon tokens


yochannah:
If we cache authentication tokens (which we do) and we get a 401 for an anonymous token - I think this happens when a token expires due to a rebuild or a certain timeout, we need to automatically delete the token and try again. Or maybe @joshkh had some session ideas about how to handle this?

yochannah:
**To test this:** 

1. Change the `:token` key for your active mine in localstorage to something wrong, perhaps "FAKETOKEN". 
 2. reload redgenes with the console open. 
 3. There should be no 401s in the console, and you should have a new token in place of your old token. If the bug isn't fixed you'll see the token set to `null` instead.

yochannah:
Wait, this is only half fixed. If a mine redeploys tokens will die, but the boot method won't happen again. How do we handle this scenario? *any* communication method could be a problem here.

yochannah:
After a brief discussion with @joshkh - we probably need to add an on-failure to im-operation, and ensure that all communication uses im-operation. 

The on-failure function should probably try renewing the authentication token and pop up a toast saying something like "There's been a communication error. Please try again. If this persists, please contact support" with a link that guides them through a cache clear and localstorage clear before contacting us... 

This is out of scope for the end-of-march  workshop target.

joshkh:
See PR #115 

This behaviour will change when we introduce log-in capabilities. For now we show a warning message to the user when a token is invalid and present the option to refresh the application.

joshkh:
![image](https://cloud.githubusercontent.com/assets/4752532/23959521/53242146-099d-11e7-8e6c-e237aabd8fbd.png)

PR #115, opened 2017-03-15
Warn users when anonymous tokens have expired.

joshkh:
This can be tested by running
```
redgenes.events.scrambleTokens()
```
from the console and then attempting to perform an operation that fetches data from the server.


joshkh:
Addresses #21

yochannah (approved):
This is jolly spiffing, old chap.

Issue #130, opened 2017-04-28
Add colouring to template screen


yochannah:
Template faceting. It would be really helpful if there was something (e.g., color coding) to indicate which template is from which "class" (e.g., Regulation, Function, Genes, Homology, etc.).  Since the template page comes up with "All" selected by default, having a color code (like you provide for faceted search results) helps a user zoom in  and say "ah, ok, the types of queries I'm interested in are under "Regulation", not "Function") and they can drill down from there.

yochannah:
![image](https://cloud.githubusercontent.com/assets/9271438/25533932/2d4b857c-2c2a-11e7-899f-1b3f058412bb.png)

Issue #126, opened 2017-04-01
IMTables Export


yochannah:
Phytomine would like to see export in IMtables

PR #135, opened 2017-05-18
Adds export, search results improvements

### LINKED ISSUE #97: Search: sometimes doesn't show anything part from the type of the result?
From @julie-sullivan 

>Disease, Ontology terms, etc.  -- can we show the name when we have it?

image

I think this might a bug of some kind as there are errors in the console? There should be more info on the screen.

yochannah:
1. Fixes #97 - search results problems in minified blueredpurplegenes. Also allows user to select multiple results and view in list result page. 
2. Adds #130 - colour tagging to templates
3. Adds export for fasta, csv, tsv to im-tables.  #126

### DIFF
--- a/less/components/search.less
+++ b/less/components/search.less
@@ -90,7 +90,6 @@
 
             .result-type {
                 line-height: 1.5em;
-                //background-color: #ccc;
                 color: @body-foreground-color;
                 margin: 0;
                 min-width: 4em;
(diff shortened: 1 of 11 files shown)


yochannah:
Test it all here, like a winner: http://devredgenes.apps.intermine.org/

joshkh:
Great job!

joshkh (approved):

PR #160, opened 2017-10-06
Add disjunction icon

harleo:
Add icon which represents a disjunction


harleo:
@yochannah 
Curious to see if the color issue presents itself in this aswell.
It propably needs viewbox adjustment, I just used the one you added for the union icon.

yochannah:
This is brilliant. @harleo you're good people. If you're interested, message me your mailing address at yo@intermine.org and I'll send over a couple of intermine stickers as thanks. :)

harleo:
@yochannah That would be too kind! Thank you :)

PR #166, opened 2017-10-24
Server changes for MyMine

joshkh:


yochannah:
Ok, merging, with the caveat that mymine is still under construction so I haven't reviewed it thoroughly - just made sure all the other bits seem to work and look ok

PR #200, opened 2017-12-13
Fix console errors - minor tidyup PR

yochannah:
### Please describe your PR:

We had a few errors popping up in the console when bluegenes loaded, mostly re-frame events/subs being called/defined in multiple places or being called when they didn't exist. This fixes most of them.


joshkh:
Visual check says yes. :)

PR #205, opened 2018-01-04
List upload

joshkh:
### Please describe your PR:

This PR introduces an ID resolver that behaves closer to the JSP version. In order to properly handle large lists and file uploads, a new resolver and its respective handlers were written. All of the components for the real-time resolver are still present but not exposed, and the plan is to merge the two together before the beta release.

Differences from JSP resolver:

* Strings and files are parsed on the BlueGenes server rather than the client. This allows us to parse large files without locking the browser.
* Users can upload multiple files, and duplicates amongst them are automatically removed

I know that in some edge cases that the result count is off for some categories, and this can be seen in the table tab headers on the Save step. This is just a matter of getting the math right based on the resolution job stats that are returned from InterMine. The actual results of the list should be accurate as seen in the list analysis page after saving.

### Review checklist: 

- [x] Region search


yochannah:
@joshkh this is brilliant! It looks nice and seems to work smoothly.

One tiny thing: when implementing css for an entire component, don't forget to sandbox the css so it doesn't leak to other components by accident, like this: https://github.com/intermine/bluegenes/blob/d53ae3e3af4f518105264ec052c8b1d02c775044/less/components/idresolver.less#L6

Issue #199, opened 2017-12-12
When loading reports after clicking on enrichment, report page fails to load.


yochannah:
Steps to reproduce:
1. Upload a list OR view a list from my data. 
2. Click on an enriched value to bring up the enrichment results in the result table
3. Click on table cell to bring up the report page for that specific item. The report page won't load successfully. 

I have no idea why this is.

PR #207, opened 2018-01-04
Use the mine name from the enrichment query when linking to report page

joshkh:
### Please describe your PR:

`flymine` was a hard coded value in the functions that add and retrieve from the enrichment history which is used when linking to the report page `/report/flymine/Gene/12345`. Now the value is dynamic.

Fixes issue #199 

### Review checklist: 

- [ ] review a _minified_ build (e.g. `lein cljsbuild min once` + `lein run`, _not_ `lein figwheel`). 

- [ ] Region search


yochannah (approved):
Perfick.

PR #214, opened 2018-01-25
Josh usecase

joshkh:
### Do not merge, this is a work in progress :)

This PR is meant to address various bugs and missing features from Rachel's use case walkthrough.

### ID Resolver
- Lists saved from the ID Resolver now appear in the My Data section

### Templates
- Optional constraints on templates can now be toggled on or off

### List Analysis & Enrichment
- Ascii < > arrows on enrichment tables have been replaced with SVG icons
- Navigating through results on the List Analysis page now works with the browser's history. This was a decently sized update but in the end it actually simplifies the plumbing of the results page (see the top of `bluegenes.sections.results.views` for comments.
  - The `:results/set-query` and its various related events for managing the history have been replaced by just two events: `:results/history+` and `results/load-history`.
  - In order to work with the browser's back button, now the app simply has to direct to `/results/[idx]` which fires the events to fetch enrichment

### Review checklist: 

- [x] Region search

### DIFF
--- a/less/components/querybuilder.less
+++ b/less/components/querybuilder.less
@@ -583,9 +583,14 @@ div.column-container {
 }
 
 .constraint-input {
+    input[type="text"] {
+        transition: .4s;
+    }
     .disabled {
         color: grey;
         fill: grey;
+        color: #666666;
+        background-color: #cccccc;
     }
 }
 //yes this is overly specific. Sorry, bootstrap is bad to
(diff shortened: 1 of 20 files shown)


yochannah:
Login doesn't work for all accounts, but that's because beta flymine isn't on the patched login branch. Everything else seems good, and the back button tested ok with enrichment click throughs! :)

PR #222, opened 2018-02-16
Use new imtables

joshkh:
### Please describe your PR:

This branch uses the newest im-tables which fixes a RACE condition where query results shown in in the results route weren't always the last query run. It also replaces the breadcrumbs with a list of recent queries and uses query titles for URL routing.

### Review checklist: 

 Before merging, confirm the following tasks have been executed

- [x] review a _minified_ build (e.g. `lein cljsbuild min once` + `lein run`, _not_ `lein figwheel`). 

Checked the following pages load results successfully and allow you to proceed to the results or report page:

- [x] ID resolver + results preview
- [x] Templates execute and show results
- [x] Templates allow you to select lists AND type in identifiers
- [x] Search (dropdown preview version)
- [x] Search (full results page)
- [x] Report page loads, including homologues and tools
- [x] Region search
- [x] Login and logout works for more than one user (use test user demo@intermine.org pw demo if needed)

This is not an exhaustive list - if you spot something else strange please bring it up!

PR #223, opened 2018-02-16
Fix enrichment race condition

joshkh:
### Please describe your PR:

Pending enrichment results are now aborted before new enrichment requests are made.

### Review checklist: 

- [x] Region search


yochannah:
Suffers from the same modal bug as #222

PR #227, opened 2018-02-27
MyMine context menus work again

joshkh:
This includes _basic_ functionality:

Right clicking a list in MyMine exposes a context menu and modals to:

1. Rename
2. Duplicate
3. Delete

I haven't introduced any additional error checking or messages, so the user isn't yet prevented from doing things like renaming a list to one that already exists. That will be round two. :)

### Review checklist: 

- [x] Region search

### DIFF
--- a/src/cljs/bluegenes/sections/mymine/events.cljs
+++ b/src/cljs/bluegenes/sections/mymine/events.cljs
@@ -93,7 +93,7 @@
   ::new-folder
   (fn [db [_ location-trail name]]
     (let [action-target (not-empty (get-in db [:mymine :action-target]))
-          uuid          (keyword (str "tag-" (str (make-random-uuid))))]
+          uuid (keyword (str "tag-" (str (make-random-uuid))))]
       (if-not action-target
         (-> db (assoc-in [:mymine :tree uuid] {:label name :file-type :folder}))
         (-> db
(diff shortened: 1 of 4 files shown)


yochannah:
I know I said no error checking was ok, but.... I was wrong. I have some thoughts. 

1. the context menu allows rename, duplicate, and delete. This is like a subsection of what's available at the top, plus rename. Why not have all on context menu _and_ all on top, maybe? (This is minor, just a bit odd. And people might not guess that rename is right click but nowhere else). 
2. A non-logged-in user shouldn't get any of these options, I think. Having the menu for rename and delete when they're impossible is a bit unfair
3. Similar to 2, I get the options to rename and delete public lists. while there _is_ a world beside them to indicate that they're public lists, it's still confusing, and the options still shouldn't be offered to the user.
4. Rename isn't accessible to a touch user even though everything else is, since it's the only item exclusively in the context menu. I don't know much about this but is it possible to launch the context menu on long press? probably? or just add to the top as per 1. above

I think fixes for 2 and 3 would be to optionally show a notice when right-clicking or hovering over buttons at the top (or tapping on buttons at the top) saying "this list is a public (read-only) list and can not be edited" or something like that.

Anyway, that's my thought ramble. I'll merge this since it _is_ an improvement upon the previous, broken context menu, but I don't think it's quite release ready yet.

PR #230, opened 2018-03-02
Pull mine details from web-properties endpoint.

yochannah:
### Please describe your PR:

Notes - still a bit rough, but feedback would be useful. 
- if the environment variables have a mine URL, it's written to index.clj as a js var so cljs can read it in. (If not present, things should still work gracefully). The purpose of this change is to allow InterMine to pass config to bluegenes about the mine running it, OR to allow bluegenes to run standalone gracefully. 
- in all cases, BlueGenes will automatically pull the mine config from the web-properties endpoint. We need to test this with older mines if possible and make sure if mines are missing config it still works gracefully or provides a meaningful error rather than fails catastrophically.

### Review checklist: 

- [x] Region search


yochannah:
updated - should work correctly even if mines.cljc was blank, so long as a URL was provided the env parameters.

PR #231, opened 2018-03-06
Fix sticky template filter container

joshkh:
### Please describe your PR:

This branch fixes the atrocious abomination that is our sticky template filter container.

### Review checklist: 

- [x] Region search

### DIFF
--- a/less/components/templates.less
+++ b/less/components/templates.less
@@ -21,7 +21,16 @@
     font-weight: bold;
 }
 
-.templates {
+
+.affix + .template-container {
+    padding-top: 102px;
+}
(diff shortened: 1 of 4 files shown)


yochannah:
Ok, merged. One note for future: it would probably be better to delete lines than just to comment them out in bulk. If we need them back, git has the history! 

I _didn't_ check the id resolver as beta flymine is down and there's a bug or unexpected input format coming from humanmine. I've made an issue about this: https://github.com/intermine/bluegenes/issues/232 - but overall I feel it's probably safe to merge anyway.

PR #239, opened 2018-03-19
Bump clj/cljs version, include highlight.js to support im-tables codegen

joshkh:
Tiny PR

* Bump CLJ and CLJS versions (better node js support)
* Add highlight.js to NodeJS dependencies

You shouldn't notice any differences between this branch and the dev branch except for maybe a node_modules directory appear in the project root when you build the app.

### Review checklist: 

- [x] Region search


yochannah:
All works ok, but the css for the top of tables is sticking off the left of the page, and the light coloured text is very hard to read:

image

Issue #283, opened 2018-10-02
Migrate from bower to npm


yochannah:
[Bower is basically deprecated these days](https://bower.io/), and all the packages available in bower.json are available on npm anyway. Let's get rid of the extra dependency. 

### Task
1. Comment on this issue to let us know you're working on it! 
1. download and install bluegenes - [guide](https://github.com/intermine/bluegenes/blob/dev/docs/getting-started.md)
2. have a quick browse of the various pages to get an idea of what they look like before we migrate the packages (they're mostly css related packages rather than functionality related packages). 
2. look at the items in [bower.json](https://github.com/intermine/bluegenes/blob/dev/bower.json) and install them via npm instead. Delete bower json. 
3. Make sure things still look the same - i.e. that we didn't break anything with the migration. 
3. Make a pull request! 

### What to do if you need help

Mention @yochannah, tweet [@yoyehudi](http://www.twitter.com/yoyehudi), pop by to say hi on [chat](http://chat.intermine.org) or if needed email yo@intermine.org <3

SergioMD15:
I'll work on this

yochannah:
@SergioMD15 Awesome! Let me know if you have any questions or issues :+1:

yochannah:
@SergioMD15 Any luck with this? :)

SergioMD15:
Sorry, I wasn't able to do it

heralden:
I just added a PR for this [=

yochannah:
I merged this too early, as @uosl pointed out - some of the scripts aren't linked properly as they'rebelow the server's public root

heralden:
I created a new PR https://github.com/intermine/bluegenes/pull/294 so hopefully this should end this chapter. [[=

PR #290, opened 2018-10-29
Migrate from bower to npm

heralden:
### Please describe your PR:

This is for fixing https://github.com/intermine/bluegenes/issues/283. I followed getting-started.md both before and after making my changes, making sure stuff looked the same. I also tested a `lein prod` production build.

I wasn't able to run the tests with `lein doo phantom test once` however, as I got the error below with a fresh repo.
`clojure.lang.ExceptionInfo: failed compiling file:<REDACTED>/bluegenes/test/cljs/re_frame_boiler/core_test.cljs {:file #object[java.io.File 0x5af5d76f ":<REDACTED>/bluegenes/test/cljs/re_frame_boiler/core_test.cljs"]}`

In addition to cleaning up `package.json` as I saw fit when merging in `bower.json`, (also creating an `.npmignore` to mimic the `ignore` field in `bower.json`) I also made changes to wherever I found `bower` to be mentioned, replacing it with `npm`.


heralden:
Thanks Travis! Looks like I didn't manage to test this properly since I hadn't cleared out the directories in `resouces/public/` after migrating to npm. Working on a fix for this now.

heralden:
@yochannah I'm not sure what is the best approach to fix this. I see now that `.bowerrc` is used to install the bower packages into `resources/public/vendor`, which some files of are imported to less source files or to the head of index.html (which is generated in `bluegenes.index`). 

`package.json` doesn't seem to have any functionality to cherry-pick packages to be moved to a different directory, so to mimic the bower behaviour, a build step has to be added to do this. A different approach would be to edit the less source files so they include stuff from the `node_modules` folder instead, and to use CDN or dist files checked into the repo, in index.html where `resources/public/vendor` is used. I see that CDN is already used for gridlex and jquery, so I'm guessing that those bower packages aren't in use.

May be some stuff I missed. What do you think?

yochannah:
hmmm, I think I'll have to pull this to have a look and a think - thanks so much for making a start! :)

yochannah:
In the process of investigating I accidentally fixed it :laughing: - editing the files to point to node_modules instead of the vendor folder. 

@uosl Thank you so much for this PR! :heart: I'll look at your other one next

heralden:
Thanks for merging!

But I noticed that some scripts in `src/clj/bluegenes/index.clj` link to files in the vendor folder, provided by bower. I think these should be replaced by CDN, similarly to how gridlex and jquery are? Since you won't be able to link to node_modules there, which is outside the web root.

yochannah:
Whoops - I totally missed that! The CDN approach sounds good.

PR #291, opened 2018-10-29
implement style guidelines checker

### LINKED ISSUE #246: implement style guidelines checker
such as https://github.com/pesterhazy/boot-fmt#configuration (which works with lein despite the name)

heralden:
### Please describe your PR:

Fixes https://github.com/intermine/bluegenes/issues/246. I first tried `lein-zprint`, but it doesn't support recursing through directories so it would either require lots of Clojure code for doing this or a shell script. (which wouldn't be multi-platform) `boot-fmt` still requires boot to be installed on your system when using with leiningen, so I didn't think it was optimal either.

I ended up settling on [cljfmt](https://github.com/weavejester/cljfmt) as the best candidate. I also added a new alias to project.clj for `lein format`, which will invoke cljfmt to fix the formatting on all Clojure source files. I ran this on the source code, and tested the newly formatted code with a `lein prod` build.


yochannah:
@uosl wow, thanks for your contributions! I've just gotten back from a few days out of the office but I will review as soon as I can :smile:

yochannah (approved):
@uosl Tested locally and everything seems to stil work as before + lein format tidies things up neatly. What would be the best way to enforce this as part of a CI pipeline?

heralden:
@yochannah Adding `lein cljfmt check` to `.travis.yml` should probably be enough. It should then be mentioned in the contributing and/or getting-started docs that people should run `lein format` before commiting.

If you mean doing the formatting automatically, I don't know if there's a way to do this as part of CI. There's git hooks that will run the formatting on the users machine (perhaps [lein-githooks](https://github.com/gmorpheme/lein-githooks)) prior to commiting. Problem with git hooks is that the feedback if something goes wrong could be lost depending on the users git client.

Personally I think the CI check plus some documentation would be a good solution since contributions will come through a PR, where they're easily spotted and fixed.

yochannah:
I think you're right - CI check + docs sounds like a good way to go.

Issue #235, opened 2018-03-13
When using an InterMine as a datasource, fail on older mines


yochannah:
I think any API version lower than 26: https://github.com/intermine/intermine/blob/gradle/intermine/webapp/src/main/java/org/intermine/web/logic/Constants.java#L212

We don't want to break public lists (possible with older APIs) or obscure features based on api version. 

I'd suggest checking this in boot.cljs as one of the first checks and displaying a friendly error explaining what is needed.

PR #293, opened 2018-10-30
Check InterMine web service version and alert when below 26

heralden:
### Please describe your PR:

Fixes https://github.com/intermine/bluegenes/issues/235. I kinda realized late that the issue said "fail on older mines", but considering my convo with @yochannah on Discord, it sounds like alerting is fine as well. [=

This PR also requires https://github.com/intermine/imcljs/pull/17 to be used together, as there was missing an endpoint for `/version`.


yochannah:
Worked exactly as expected, nice one! :fireworks:

PR #303, opened 2018-12-06
Save the server! Fix for Region validation.

### LINKED ISSUE #23: Region search: Validate input.
Right now whatever we input into the box is submitted to the sever. When the input is invalid we don't get verbose or meaningful error messages back, so we should validate the regions clientside before submission.

Demonstrative example: 
example region search input

The 'ss' appended to the end of the input in that screenshot results in a 500 from the server, and no results for any of the regions.

sehgalvibhor:
### Please describe your PR:
Closes #23 
Disables the button if the user does not follow the syntax for the input. This might be a sweet temporary fix, can be improvised in the near future based on subsequent discussion/review.

## Review checklist: 

- [x] review a _minified_ build (e.g. `lein prod`, _not_ `lein figwheel`). 

- [ ] Region search

### DIFF
--- a/src/cljs/bluegenes/pages/regions/views.cljs
+++ b/src/cljs/bluegenes/pages/regions/views.cljs
@@ -137,6 +137,17 @@
      [:div.feature-tree-container
       {:class (if @results "shrinkified")} [feature-types-tree]]]))
 
+(defn matches-regex?
+  [v regex]
+  (boolean (re-matches regex v)))
+
+(defn is-valid?
+  [v]
+  (if (nil? v)
+    false
+    (let [regexp (clojure.string/join "--" (split v #"\n"))]
+      (matches-regex? regexp #"([A-Za-z0-9\.]+(:)[0-9]+((\.\.)|-)[0-9]+(--))*([A-Za-z0-9\.]+(:)[0-9]+((\.\.)|-)[0-9]+)"))))
+
 (defn input-section
   "Entire UI input section / top half of the region search"
   []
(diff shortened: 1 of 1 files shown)


src/cljs/bluegenes/pages/regions/views.cljs
```diff
@@ -137,6 +137,17 @@
      [:div.feature-tree-container
       {:class (if @results "shrinkified")} [feature-types-tree]]]))
 
+(defn matches-regex?
+  [v regex]
```
yochannah, line 141: The regex works great! Could we add a docstring here explaining what the regex is checking for, in plain english? You can see an example of a docstring on line 152 below.

yochannah (changes requested):
Hey @sehgalvibhor! Sorry for taking so long to test this one. The code you've implemented works great, but I think it needs a little bit more before we can merge it - my concern is that users who are trying to use the region search might not quite know _why_ the form is disabled, in its current state. It might be easier for them if we provide some feedback - perhaps a polite error message explaining the mandatory format and telling them which line is problematic, if there are multiple lines.

One option to implement this might be custom html5 form validation: https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation#Customized_error_messages but other reasonable implementation methods can be considered as well.

If you want to read more about feedback on errors: https://www.nngroup.com/articles/error-message-guidelines/

sehgalvibhor:
Got it! I'll make the necessary changes and update the PR.

yochannah:
Hey @sehgalvibhor! This PR has been around a while; I'm going to close it for now but if you wish to work on it further please feel free to re-open it.

PR #310, opened 2018-12-25
Sort by rank - ascending order

### LINKED ISSUE #129: Sort templates by rank
Courtesy of @julie-sullivan :

"I just added rank to the template JSON"

```
"name":"AllGene_Chromosome",
         "comment":"",
         "tags":[  
            "im:aspect:Genes",
            "im:public"
         ],
         "rank":"2",
```

sehgalvibhor:
### Please describe your PR:
Closes #129 

### Review checklist: 

- [ ] review a _minified_ build (e.g. `lein prod`, _not_ `lein figwheel`). 

- [ ] Upload page resolves IDs as expected
- [ ] Region search

### DIFF
--- a/src/cljs/bluegenes/pages/templates/subs.cljs
+++ b/src/cljs/bluegenes/pages/templates/subs.cljs
@@ -63,7 +63,7 @@
                        (if category
                          (some? (some (fn [tag] (filter-pred tag category)) (:tags details)))
                          true))]
-     (filter (partial template-contains-string? text-filter) (filter filter-fn templates)))))
+     (sort-by (comp :rank second) < (filter (partial template-contains-string? text-filter) (filter filter-fn templates))))))
 
 (reg-sub
  :selected-template-name


yochannah:
@sehgalvibhor I'm really sorry but can't actually get this to work. I see this order, for HumanMine (tested with Fly too). 

image

But on HumanMine the order is 

image

I wasn't sure *which* was the correct order - it's conceivable that HumanMine is wrong. This script also matches humanmine's ordering, though https://codepen.io/yochannah/pen/ebGZRp?editors=0011#0

sehgalvibhor:
Okay, must be something wrong I did. But, a part of the result is sorted and a part is not. 
https://jsoneditoronline.org/?id=599c446840ed40dc9fa49d1bc9c94a64

Maybe this output will help (Flymine)! If you see, a lot of elements are sorted on the rank. Interestingly, when I change rank to name, it does sort everything alphabetically.

yochannah:
hmm, at a guess, I think could be because the rank is (somewhat incorrectly) returned as a string. 

```clojure
(.parseInt js/window "12")
```

that's the basic syntax (using js methods) to parse a string to an int.

yochannah:
I've logged the issue here: https://github.com/intermine/intermine/issues/1976

sehgalvibhor:
I also think that's the case, thank you so much :)

yochannah:
Okay! As a side note, we should probably just parse the strings to int for now rather than wait for InterMine.

sehgalvibhor:
Okay, let me try typecasting to see if it works correctly or not.

sehgalvibhor:
Um, I am directly applying type casting in the comp function ``` (comp #(.parseInt js/window (:rank second))) ```. Is this wrong way of doing it, as the result is not sorted?

yochannah:
Whoops, I just hopped onto this PR to see what the status was and realised it was waiting on my response. So sorry!!

yochannah:
The ticket to return ints as ints ( https://github.com/intermine/intermine/issues/1976 ) has been closed as a won't/can't fix, so we definitely need to work around the issue. I suspect the reason the comp above wasn't working is that we had a lot of NaNs - there are two types of rank, an int presenting as a string, e.g. `"12"` and also real strings, for templates with no rank yet - `"unranked"`. I'll merge your PR but I've added a little if function to turn unranked templates into rank 99999 as a hack to make them sort to the end.

PR #332, opened 2019-04-01
[WIP] Add support for new config style

### LINKED ISSUE #288: ID resolve default config - support old and new config style
The ID resolver component uses the web-properties endpoint to fetch defaults for the ID resolver, among others things. 

The bag endpoint results used to look like this: 

```json
    "bag": {
      "results": {
        "table": {
          "size": "5"
        }
      },
      "example": {
        "identifiers": "CG9151, FBgn0000099, CG3629, TfIIB, Mad, CG1775, CG2262, TWIST_DROME, tinman, runt, E2f, CG8817, FBgn0010433, CG9786, CG1034, ftz, FBgn0024250, FBgn0001251, tll, CG1374, CG33473, ato, so, CG16738, tramtrack,  CG2328, gt"
      }
    }
```

Note `bag.example.identifiers`. Newer versions of the config look like this: 

```json
    "bag": {
      "results": {
        "table": {
          "size": "5"
        }
      },
      "example": {
        "identifiers": {
          "default": "PPARG, FTO, 3949, LEP, 946, MC3R, 9607, LPL, LDLR, P55916, 335, GLUT4, Notch1, SLC27A1",
          "protein": "P37231, PPARG_HUMAN"
        }
      }
    }
```
We now have bag.example.identifiers, with subchildren `default` (this maps to Gene for legacy reasons) and `protein`. This is potentially infinitely extensible to however many child class properties we wish. 

Right now BlueGenes only supports the first option I think. We should support both.

The relevant file to change is here: https://github.com/intermine/bluegenes/blob/2eeb54e5ab2d03f32802593b2a90ba6dee4c4749/src/cljs/bluegenes/events/webproperties.cljs#L16 where we store the exaample to the app-db.

allentiak:
### Please describe your PR:

Closes #288.

For now, this is a Work In Progress.

### Review checklist: 

- [ ] review a _minified_ build (e.g. `lein prod`, _not_ `lein figwheel`). 

- [ ] Upload page resolves IDs as expected
- [ ] Region search

### DIFF
--- a/src/cljs/bluegenes/events/webproperties.cljs
+++ b/src/cljs/bluegenes/events/webproperties.cljs
@@ -13,18 +13,14 @@
    :regionsearch-example         (get-in web-properties [:genomicRegionSearch :defaultSpans])
    ;;this needs to be passed in as an arg or pulled from the branding endpoint.
    :icon                         "icon-intermine"
-   :idresolver-example           {:Gene    (get-in web-properties [:bag :example :identifiers])
-                                  ;; there should be details implemented for non-gene ID
-                                  ;;defaults,
-                                  ;; but we commented out the service for some reason
-                                  ;; on intermine's side.
-                                  ;; hopefully, it will return.
-                                  ; :Protein "Q8T3M3,FBpp0081318,FTZ_DROME"
-}
-;   :default-query-example        {;;we need json queries to use the endpoint properly
+   :idresolver-example-default    {:Gene (get-in web-properties [:bag :example :identifiers :default])}
+   :idresolver-example-protein    (get-in web-properties [:bag :example :identifiers :protein])
+   ;:default-query-example        {;;we need json queries to use the endpoint properly
                                   ;;https://github.com/intermine/intermine/issues/1770
                                   ;;note that the default query button won't appear
                                   ;;until we fix this issue
+                                  ;;UDPATE: #1770 has been closed.
+                                  ;;TODO: fix this.
                                 ;}
 })
 


allentiak:
@yochannah 
What do you think about this? Am I on the right track?

yochannah:
Almost! 

Consider http://www.flymine.org/flymine/service/web-properties vs http://www.humanmine.org/humanmine/service/web-properties - we need to support *both* styles of json, where one property is more nested than the other. 
We also probably shouldn't hardcode to Gene and Protein in the key names if possible - while this will work for many intermines, it won't work for all of them (some might have different favourite class types), and we should fail gracefully if a value *isn't* present.

re ticket 1770 - that is probably a separate issue :)

yochannah:
@allentiak I'm working on some related issues on the id resolver at the minute. Would you mind terribly if I scooped this issue and implemented the solution? (If you'd rather keep on, that would be fine too).

allentiak:
Could you please give me a little bit more time? I plan to tackle this
tomorrow...
I only have one question: I was thinking to approach this using
clojure.core/some. What do you think about it?
https://lambdaisland.com/episodes/clojure-core-some

yochannah:
Seems worth a try! I can wait :)

allentiak:
I am setting up a local dev environment which accesses remote data. The documentation is not straightforward regarding this. I will first update it and open a PR for that.

allentiak:
I am having some problems setting up a local development environment which accesses remote data. The documentation regarding this is scattered. Before I forget how to go about it, I will improve the documentation and open a PR for that.

allentiak:
> I will improve the documentation and open a PR for that.

See #339

yochannah:
Reviewing old PRs - I think this issue has been fixed and the PR is aged; closing for now.

PR #337, opened 2019-04-29
Remove terrible inline styles and font icons.

yochannah:
…css is good

### Review checklist: 

- [x] review a _minified_ build (e.g. `lein prod`, _not_ `lein figwheel`). 

- [x] Upload page resolves IDs as expected
- [x] Region search


yochannah:
Report page button isn't checked - there's a bug in the homologue display. I think this should be a Tool API tool anyway, and it's been buggy for a while, so I'll merge.

Issue #267, opened 2018-08-23
full search is buggy


yochannah:
Comments from @sergiocontrino 

- Search (full results version): looking for eve gives a few categories in fly, anatomyterm and interactionexperiment claim 2 hits, but none is shown when selected
- Search needs reload to work!

This is one of the oldest bits of the codebase and may need serious refactor

yochannah:
>Search (full results version): looking for eve gives a few categories in fly, anatomyterm and interactionexperiment claim 2 hits, but none is shown when selected 

- still present

PR #355, opened 2019-06-11
Fix mines not initializing properly from local storage

### LINKED ISSUE #354: Mines don't initialize properly after page reload
Something bad happens when loading from data saved to local storage!

This manifests itself most prominently as the search bar refusing to receive input, as reported in issue #267.

heralden:
Fixes #354 and *partly fixes* #267.

I removed some code/functionality I thought no longer applied since we now have an InterMine registry, but this may have resulted in bugs in use cases I'm not aware of. So if you know of any use cases to test it, please do! (I've only tested starting from a fresh instance, starting from local storage, and with a mine selected. I have not tested when coupled with InterMine, but it *should* work as I have accounted for it.)

One thing we might wish to implement as a separate PR: (I'll gladly create an issue for this)
- If :current-mine is no longer present in :registry, it should change :current-mine to :default and delete the old :current-mine from :mines (this was previously handled in the :boot fx, but since we've replaced the local config with the registry it will have to be handled later on)

### DIFF
--- a/src/cljc/bluegenes/mines.cljc
+++ b/src/cljc/bluegenes/mines.cljc
@@ -1,7 +1,0 @@
-
-(ns bluegenes.mines)
-
-(def mines {:default        {:id         :default
-                             :name "FlyMine"
-                             :service    {:root "flymine.org/flymine"
-                                          :token nil}}})
--- a/src/cljs/bluegenes/db.cljc
+++ b/src/cljs/bluegenes/db.cljc
@@ -1,9 +1,8 @@
 (ns bluegenes.db)
 
 (def default-db
-  {;;events/boot.cljs auto-selects the first mine available if current-mine
-   ;;doesn't exist for any reason.
-   :current-mine :flymine-beta
+  {:current-mine :default
+   :fetching-assets? true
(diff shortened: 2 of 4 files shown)


heralden:
Something to note when you look into this PR: You might have to include #356 for it to work. I thought they'd work independently of each other when looking at the diffs, but that doesn't seem to be the case. It would probably have been easier if I kept them as one PR, but I thought they fixed two different issues. §=

heralden:
Fixed what I mentioned in the previous comment with the latest commit.

yochannah:
@uosl  looks good from an inspection of the code and running in prod mode. Could I get you to run `lein format`? Once the build passes I think I'm happy to merge.

heralden:
Done! I'll try and remember to format next time. (I think we have an issue for setting up a git-staged hook for doing this automatically?)

yochannah:
@uosl Looking at the new travis failures, it looks like if you delete localstorage, the initial boot process fails because it's hitting a 301 redirect. I think the fix might be just to add www. to the default flymine URL?

yochannah (approved):

PR #356, opened 2019-06-11
Fix load-other-mines/fetch-registry not running

heralden:
I noticed that `load-other-mines` was not running since the namespaced keywords were incorrect. I fixed these (in the same vein as the keywords used in *events/webproperties.cljs*) and also renamed it to `fetch-registry` so it would be consistent with the naming scheme of prefixing `success-` to the fetch keyword.


heralden:
Correction: There was nothing wrong with the namespaced keywords! (I just learned now that `::some-ns/keyword` is valid when you've required the ns using `:as`.)

I must have messed up when I was testing the fx, cause testing from the dev branch now; it definitely fires.

I will close the PR since it doesn't do anything after all.

PR #358, opened 2019-06-12
Allow symlinks when serving bluegenes-tool-path

### LINKED ISSUE #348: Local development bluegenes tools: does the npm link trick really work?
When I tried to serve a dev tool locally using `npm link`, I get a 500 trying to serve those files. The same files physically there (rather than symlinked) works fine. I like to imagine I tested this feature before putting it in the tutorial - https://github.com/intermine/bluegenes/blob/dev/tools/docs/tools.md#development-tools - but maybe I didn't? 

It would be nice for someone who is not me to try the following: 

1. generate a basic tool following https://github.com/intermine/bluegenes/blob/dev/tools/docs/tool-api-tutorial.md - or just clone [an existing tool](https://github.com/intermine/bluegenes-tool-cytoscape) to a directory that's *not* the bluegenes tool directory
2. try running npm link to set up your bluegenes tools environment, as per https://github.com/intermine/bluegenes/blob/dev/tools/docs/tools.md#development-tools

tell me if the tool files load on a report page, or if they 500 like mine did! :)

heralden:
Fixes #348. This was a bit tricky to test, until I realised the browser was reading from disk cache and switched to testing with curl instead. [=

Compojure's `files` route doesn't follow symlinks unless you add `:allow-symlinks? true` to the option map.

### DIFF
--- a/src/clj/bluegenes/routes.clj
+++ b/src/clj/bluegenes/routes.clj
@@ -25,7 +25,7 @@
              ;; when running uberjar or clojar targets,
              ;; and make the jars about a million megabytes too big.
 
-  (files "/tools" {:root (:bluegenes-tool-path env)})
+  (files "/tools" {:root (:bluegenes-tool-path env), :allow-symlinks? true})
 
   (GET "/version" [] (response {:version "0.1.0"}))
 


yochannah:
>Compojure's files route doesn't follow symlinks unless you add :allow-symlinks?
true to the option map.

Ah-hah!

yochannah (approved):
Works like a charm :+1: tested on my local machine. Nice one!

yochannah:
The build failure is just a docs issue and has been fixed on master. Merging.

PR #359, opened 2019-06-13
Externalise tool API server and import route from dependency instead

### LINKED ISSUE #327: migrate tool api server to its own package
The tool api server will also need to be used by the InterMine Creator, for the tool selection step. It won't be too hard to migrate it to its own package, and package it up as a JAR so that @julie-sullivan can use it to serve the tools to the creator.

heralden:
Resolves #327. Looks like it was a simple matter of using `compojure.core/routes` to combine the route from the `bluegenes-tool-store` package, and it worked just as if it were included in `bluegenes`!

I had to update ClojureScript dependency as well since otherwise it
would trigger a `transit-java` bug due to an older version being pulled:
`java.lang.NoSuchMethodError: com.cognitect.transit.TransitFactory.writer`


yochannah (approved):
Works perfectly- tested the api route, the report page, in both prod and dev build modes.

yochannah:
@uosl I would like to merge this, but (my fault) it fails the formatting validation since I used the GitHub UI to resolve a merge conflict - and I can't seem to push to your branch. I assume the "allow maintainers to push edits" button is still checked? Alternatively, could you pull this, run lein format, and push again?

heralden:
@yochannah  "Allow maintainers" should be checked. I can resolve this like you describe, but just want to check with you first: do we still need this if we're not going to use the Tool server in the cloud wizard after all?

yochannah:
hmmmm. I think there might be value in leaving it separate, as we could end
up running a tool server that others use if they wish, and it might be nice
to be able to maintain that separately from bluegenes itself. I'm not
strongly sold on this idea, though - do you have a preference?

heralden:
I think that's a reasonable argument. I'll go get it merged!

PR #360, opened 2019-06-13
Replace use of npm link in docs to use path instead as it's simpler

heralden:
As we discussed; npm installing a filesystem path is simpler since it's a single step process instead of two. [=


yochannah (approved):
👍

yochannah:
The build error is fixed on master so I'm merging this even though it failed the travis check.

Issue #362, opened 2019-06-18
Rework routing and URL paths


heralden:
Here's the plan so far, as discussed with @yochannah.

## URL paths
- Gene
old: /reportpage/humanmine/Gene/1208821
new: /humanmine/report/Gene/1208821
- Organism
old: /reportpage/humanmine/Organism/1208821
new: /humanmine/report/Organism/1208821
- Lists
old: /results/PL_Pax6_Targets
new: /humanmine/list/PL_Pax6_Targets
- Query
old: /results/Custom Built Query
new: /humanmine/list/Custom Built Query
- Search
old: /search
new: /humanmine/search?key="gata1"

## Goals
- [x] Rename reportpage path
- [x] Query string including keyword for *search* path
- [x] Replace *results* path with *list*
- [x] If a list is not in the history, automatically query for it (such as the case when directly opening path `/humanmine/list/PL_Pax6_Targets`) #66
- [x] Make sure report page deep links work from a cold start in a different mine #78
- [x] Deep linking of mines (try to have the mine namespace optional in the URL path, but fallback to specifying `default` if this isn't practical) #352
- [x] Replace secretary+accountant with reitit for reverse-routing (so we can easily rename routes)
- [x] Use anchor elements and href (to allow *open in new window*) #90
- [x] Use HTML5 history API instead of hash history
- [x] Make sure back button works with the new solution

Issue #352, opened 2019-06-04
deep link to an intermine using its registry namespace


yochannah:
Something like 

http://localhost:5000/mine/flymine 

Might make sense.

Namespaces come from registry.intermine.org.

Issue #90, opened 2017-02-27
hyperlinks: allow them to be opened in new windows


yochannah:
From Gos: 

Command-click on Mac doesn't open new tab but instead follows the link from front page to
view list.  This seems true throughout - bit frustrating as I like to shift-click on lots
of things and then go through the tabs.

yochannah:
I think the fix for this is to add correct href attributes to all links

Issue #78, opened 2017-02-16
Report page deep links don't work from a cold start in a different mine


yochannah:
If your default / current mine is flymine beta, this link won't work: 
http://devredgenes.apps.intermine.org/#/reportpage/humanmine-beta/Gene/1239611 

You may get something like this:

![image](https://cloud.githubusercontent.com/assets/9271438/23020597/b2b6a26e-f43f-11e6-802f-8393d043f2dd.png)

Basically we should make sure deep links don't depend on the active app state matching the mine name in the link.

When addressing this also make sure the system works for list links. see #66

PR #372, opened 2019-06-28
Rework routing and URL paths

heralden:
This PR is paired with #362. Basically it changes how routes are handled (as well as library used), adding numerous features/fixes along the way to make sure it was a good choice. [=

Here's a copy of the list of goals and related issues:

- [x] Rename reportpage path
- [x] Query string including keyword for *search* path
- [x] Replace *results* path with *list*
- [x] If a list is not in the history, automatically query for it (such as the case when directly opening path `/humanmine/list/PL_Pax6_Targets`) #66
- [x] Make sure report page deep links work from a cold start in a different mine #78
- [x] Deep linking of mines (try to have the mine namespace optional in the URL path, but fallback to specifying `default` if this isn't practical) #352
- [x] Replace secretary+accountant with reitit for reverse-routing (so we can easily rename routes)
- [x] Use anchor elements and href (to allow *open in new window*) #90
- [x] Use HTML5 history API instead of hash history
- [x] Make sure back button works with the new solution

Something I haven't tested is the creation of lists through saving a query, set operations, etc. Although they should *probably* work as `:results/history+` which they rely on still has the same behaviour.


yochannah:
@uosl - I'm taking a look at this now, but can I get you to make sure travis passes before we merge?

uosl:
The Cypress test failures are dependent on this fix: https://github.com/metosin/reitit/pull/301

src/cljs/bluegenes/core.cljs
```diff
@@ -26,22 +25,24 @@
...
                   (ocall js/document "getElementById" "app")))
 
-(defn navigate-to-deep-links []
-  (let [url (oget js/window "location" "hash")
-        hashless-path (last (clojure.string/split url #"#"))]
-    (cond (> (count url) 2) ;; if there is more than #/ in the url, navigate there
-          (if (= (first hashless-path) "/")
-            (navigate! hashless-path)
-            (navigate! (str "/" hashless-path))))))
+;; TODO replace or delete this function depending on needs with reitit
```
yochannah, line 33: what are the considerations here with regards to the todo?

src/cljs/bluegenes/events.cljs
```diff
@@ -60,24 +60,52 @@
...
+                                     :name (:name mine-m)
+                                     :id mine-kw}))))
+
+;; This event handler doesn't do anything, as it exists only for `async-flow` to
+;; observe so it knows that `:init-mine` has successfully completed.
+(reg-event-fx
+ :success-init-mine
+ (fn [_ _] {}))
+
+;; This is for setting :current-mine based on the URL path. As opposed to
```
yochannah, line 80: nice comments here!

src/cljs/bluegenes/core.cljs
heralden: Deep linking seems to be working fine without this, so we can delete this code.

Edit: I can see it's specifically for handling links with the hash router, which we don't use anymore.

src/cljs/bluegenes/route.cljs
```diff
@@ -0,0 +1,230 @@
...
+  (:require [re-frame.core :refer [subscribe dispatch reg-sub reg-event-db reg-event-fx reg-fx]]
+            [oops.core :refer [ocall]]
+            [reitit.core :as r]
+            [reitit.coercion :as rc]
+            [reitit.coercion.spec :as rcs]
+            [reitit.frontend :as rf]
+            [reitit.frontend.controllers :as rfc]
+            [reitit.frontend.easy :as rfe]))
+
+;; # Quickstart guide:
```
yochannah, line 11: This is fabulous, thank you!

yochannah:
perfect, thanks

src/cljs/bluegenes/route.cljs
```diff
@@ -0,0 +1,230 @@
...
+
+(reg-event-fx
+ ::navigate
+ (fn [{db :db} [_ route & [params query]]]
+   {::navigate! {:k route
+                 :query query
+                 :params (update params :mine #(or % (:current-mine db)))}}))
+
+(reg-event-db
+ ::navigated
```
yochannah, line 44: could we add a comment about when/why this is used?

heralden: Is it worth it? It's pretty internal to how the router works, so it's not that important to know how it works, and those who do want to know can read it as it's still a pretty short and simple function. [=
I at least can't think of a succinct and useful comment to put there.

yochannah:
"you probably won't need to use this as it's an internal helper method" ?

src/cljs/bluegenes/route.cljs
heralden: >"you probably won't need to use this as it's an internal helper method" ?

Sure, marking it as internal sounds reasonable.

src/cljs/bluegenes/route.cljs
```diff
@@ -0,0 +1,230 @@
...
+      [{:start #(dispatch [:set-active-panel :mymine-panel])}]}]
+    ["/report/:type/:id"
+     {:name ::report
+      :controllers
+      [{:parameters {:path [:mine :type :id]}
+        :start (fn [{{:keys [mine type id]} :path}]
+                 (dispatch [:set-active-panel :reportpage-panel
+                            {:type type :id id :mine mine}
+                            [:load-report mine type id]]))}]}]]])
+;; You can do initialisations by adding a :start function to :controllers.
```
yochannah, line 203: I'm not quite sure what you mean here by teardowns and initialisations here, or why I might want to use them.

yochannah (approved):
:medal_sports:

src/cljs/bluegenes/route.cljs
heralden: Most of the routes have a `:controllers` vector with a `:start` map. While you can copy-cat the existing `:start` maps when you create new routes, there aren't any using `:stop` (teardowns) currently. I wanted to make sure to write it somewhere so people know they exist as well.

PR #373, opened 2019-07-11
Cleanup old phantom, update docs and cypress

heralden:
I don't think PhantomJS for testing is used anymore?

I updated the docs for how to run the Cypress tests, and removed old references to the PhantomJS tests. I also updated the Cypress dependency to the latest version.


yochannah (approved):
as usual your commits are always great :)

PR #374, opened 2019-07-12
Properly handle switching mines and list assets from other mines

### LINKED ISSUE #370: Nameless lists appear when switching mines
Reproduction: Start from a cleared local storage, change to *humanmine*, then open *My Data*. All the lists should have blank names. If you refresh, you should only have 4 nameless lists. If you clear local storage and refresh again, everything should look like normal, with 0 nameless lists.

heralden:
Fixes #370.

Note that this PR builds on #372, (adds the last two commits) so that should be merged before this!


yochannah (approved):
successfully reproduced the issue and confirmed this fixes it :100:

yochannah:
Note: the failing travis check was introduced on a previous merge (that was *also* passing until we merged it, what?!) and not the result of code in this PR.

Issue #363, opened 2019-06-25
remove old postgres and mymine tags infrastructure. 


yochannah:
These are relics from a forgotten, less civilised age when we thought people might have BlueGenes accounts (they won't, only InterMine accounts). 

We should strip out references to tags in mymine, and remove Postgres. There's no need for it.

heralden:
Closing this as it got fixed in #378.

PR #378, opened 2019-07-22
Improve tool API support

heralden:
At the point of PR creation, support for displaying tools on the list page has been implemented. (Tools are also properly filtered on the list or report pages based on what they support.)

I might add some other tool API improvements here as well.


heralden:
I added commits that fix #363.

I tried my best to disentangle the list-related code which we wanted to keep, from the tags-related code we wanted to remove. Nothing has been affected as far as I can tell, but you should probably test as well.

yochannah (approved):
- [x] Read through the files - looks _great_. Tidy and easy to understand what's going on. 
 - [x] Tested the new navigate link and it does what you'd expect.
 - [x] tested on minified `lein prod` version.

Issue #382, opened 2019-08-01
Tool API spec update: don't show tools that have no data


yochannah:
Some tools, such as the Protein Atlas expression tool, only work with certain data sets from certain mines. See https://github.com/intermine/bluegenes-protein-atlas-tissue-expression-visualizer/blob/master/src/queryData.js#L6 for an example 

- [ ] We should figure out how to make bluegenes look for certain data entities in the model and only try to load the tool if it is compatible with a given intermine.
- [ ] release version 1.1 of the spec
- [ ] make sure tools can specify which version of the spec they comply with
- [ ] update all tools, yowza.

heralden:
Looks like we can confirm whether the mine model supports it by checking the keys in `:assets :summary-fields mine` in appdb.
Should we simply add a field to *config.json*, perhaps called `model`, which takes an array of these strings? (eg. `["proteinAtlasExpression", "atlasExpression"]`)

yochannah:
This sounds right! It should probably be full paths, though - e.g. `Gene.proteinAtlasExpression` or `ProteinAtlasExpression`

heralden:
I ended up calling it `depends`, and with the elements specified as `ProteinAtlasExpression` like you suggested.

It will be a non-breaking change, since the tools that need this feature can add the *depends* key while those that don't will work like before. Do we need to version the tool api in that case?

yochannah:
I love non breaking changes!!! Let's version it anyway - semver principles
suggest this should be a minor version or even a patch since it's
non-breaking https://semver.org/

heralden:
According to semver we should be at 2.2.0.
Since there have been three changes to the tool api, from my recollection:
- Removing support for formats "list" and "lists" (definitely a breaking change)
- Adding depends key (not a bug fix, but a backwards compatible feature)
- Navigate function (another backwards compatible feature)

We only have one tool that used "list", (https://github.com/intermine/bluegenes-GO-term-visualizer) which I'll fix once I can fix https://github.com/intermine/imjs/issues/52.

Now I have some questions. How are we going to release version 2.2.0 of the tool api? Is it enough to hard-code the version number somewhere in the bluegenes tool code, or should we also keep a changelog and link to it in the docs?

To do it properly, we would also have to update the *config.json* of all existing tools (and the tool generator) to specify `apiVersion: 2`, as any missing the key would be interpreted as supporting only `1`. (Obviously it would only be necessary to specify the major version number, since the other two are non-breaking.)

[...]

Issue #346, opened 2019-05-21
tool api page - show tool path


yochannah:
it would be nice to see to tool path being used by bluegenes somewhere in the admin ui.

PR #383, opened 2019-08-05
Interactive tool store using backend API for npm

### LINKED ISSUE #170: tool selector page - "app store"
This is a page for BlueGenes admins to browse through available tools and enable or disable them for their Mine.

Will require setup to write config to server. Flat edn config file? db? edn probably easier.

### LINKED ISSUE #361: better tool folder management for dokku
Currently we enter the docker container and npm install the tools: https://github.com/intermine/bluegenes/blob/dev/docs/tools/dokku-tool-installation.md 

I have a feeling this might nuke the tools every time we redeploy. What is a better solution? 

- Figure out how to specify a tools folder outside container? Will that allow updates to the external folder that automatically reapply to bluegenes? I think dokku allows us to link eternal volumes somehow
- script npm installing everyhing in the tool folder's package json, post-install. This has the advantage of being low effort, and we could still enter the container if we want to install a tool without redeploying bluegenes. 
- other, smarter idea?

heralden:
Uses https://github.com/intermine/bluegenes-tool-store/pull/1 to implement an interactive tool store that allows you to manage tools via the new backend API for npm.

Since we currently don't have authentication for this, we probably shouldn't merge this until this is in place.

Note that this uses the dependency `[org.intermine/bluegenes-tool-store "0.2.0-ALPHA"]` which isn't on Clojars. You can acquire it locally by running `lein do pom, jar, install` in the checked out [bluegenes-tool-store branch](https://github.com/uosl/bluegenes-tool-store/tree/feature/npm-tool-api), then restart bluegenes making sure to include `:enable-unsafe-routes true` in your bluegenes environment.

This should close #361 and close #170 once it's finished.

2019-08-05-160408_863x970_scrot

Here's a TODO list for this PR:
- [x] Authentication so that only the admins for a mine can manage its tools (dependent on in-progress work at Intermine) https://github.com/intermine/intermine/issues/2099
- [x] ~~A third category of *Unsupported tools* which are incompatible with the mine's data model~~ I ended up adding an alert displayed in the *box* for these tools. Since the tools could well work for a different mine from the mine switcher, we shouldn't disencourage mine maintainers from installing it just because it won't work for their mine. #382 
- [x] Make `tools/package.json` minimal (only include `dependencies` key) and remove it from version control. (Make sure to have code in bluegenes-tool-store that initializes the file if it doesn't exist, before running npm commands. In addition, all tools should be installed by default on first startup.)
- [x] Show tool API path in bluegenes web debug console #346 
- [x] Respond with updated tool list to `/api/tools` to save the extra request from the client getting the updated list
- [x] Always use `:default` mine service when querying tool API (as other mines would fail the privilege check)
- [x] When first initialised, only install all npm packages with the *bluegenes-intermine-tool* tag that are under the intermine user (others have to be added with the tool store; reasoning being to avoid maliciously created npm packages)
- [x] Handle case where it's an older Intermine without support for checking superuser flag (in this case we have to remove the interactive functionality, replacing them with the npm commands you can run manually instead along with dir name, and a note that upgrading to Intermine 5+ will make this automatic)


yochannah:
@uosl I note that has moved to "to be tested", but still has WIP in the title - is it ready for review?

heralden:
This one is ready for review: https://github.com/intermine/bluegenes-tool-store/pull/1
Which I guess includes this one. I left the WIP in the title since **this PR** should not be merged before the bluegenes-tool-store PR is merged and the dependency updated.

yochannah:
I love how smooth this is! I can confirm that for an older InterMine, no actions are allowed to be taken since there's no superuser flag. 


Some small design questions while I'm setting up biotestmine to test this with the superuser flag switched on: 

(I'd suggest reading this: https://www.nngroup.com/articles/confirmation-dialog/ before reading the next few points for context on good usability practice)

- The "Remove" button on each tool - it just straight-away deletes stuff. In some cases I'd suggest adding a "please confirm you meant to do this" button - and for uninstall-all I'd suggest we _do_ do this - but I think uninstalling and reinstalling is so easy, it maybe doesn't matter if we do it by accident. We don't really have any support for custom config that might get lost due to an incautious / accidental uninstall. Does that sound about right? 
- "Update all" - I don't think this is destructive so no need for confirmation

And these I've not yet fully explored so don't know if they exist: 
- "Remove all" - probably should exist if it doesn't, and probably _should_ have a "This will uninstall all of your tools [Don't uninstall] [I understand, uninstall all tools]" confirmation dialogue - but this is a weird case so low priority. Perhaps we should make it as an issue unless you're bored and fancy adding it to this PR... what do you think? 
- "Update to v1.0.5" (or whatever version) - just one tool. I don't think this needs an "are you sure" dialogue either.

heralden:
I agree that a confirmation dialog is unnecessary since none of these actions are dangerous. The closest thing would be an *Uninstall all* button, but I refrained from implementing one specifically for this reason (and I felt it unnecessary for the first version).

Similarly, while I felt being able to update specific tools to specific versions would be useful at some point, there's even more work needed to support that so I thought it was of a low enough priority to be added further down the road.

In conclusion, should we create issues for these two things and go on with merging this PR as it is? ~~(Once I've updated bluegenes-tool-store, which I'll do now.)~~
Update: This PR is ready to be merged now.

yochannah:
based on the comments from @uosl, I"ll merge this now :)

Issue #212, opened 2018-01-23
Logon: User shouldn't appear to stay logged in when switching mines


yochannah:
currently if user loads bluegenes and logs on to flymine then switches to mousemine they shouldn't appear to still be logged in.

PR #384, opened 2019-08-06
Fix expired token alert not dismissed on refresh button or mine switch

### LINKED ISSUE #133: expired token notification - remove if mine is changed
To reproduce: 

1) Have an expired API token or spoof it programmatically
2) Decide to change mine rather than refresh tokens because I dunno, seemed like a good idea
3) I'm in a new mine now with a good token but that notification won't go away gosh darn it! I feel like it should have gone away when I changed mines.

I don't think we should just add an "x" dismiss button to the notification - dismissing the notification and finding everything is broken isn't very useful...

heralden:
Also remove scramble-token debug code and make sure `:reboot` is used
instead of `:boot`. Closes #133.


yochannah (dismissed):
Hmm, as part of testing this I discovered #387 and #388 but I can verify that these errors are also present in dev, so that's no reason to stop the merge. 

One thing I did note is that the undismissable dialogue is still present when I am logged in to an InterMine. I think we should fix this before merging.

yochannah:
+ It might make sense to leave scramble token in, since it's handy :laughing:

heralden:
The last commit fixes #212, #387 and the persistent dialog when logged in.

Please let me know if you manage to break this as well!

heralden:
Forgot to add reference that 8a5a25180bd506f3e3cf57dde494f77adf032c3f closes #388.

heralden:
If the CI fails, this would be because Flymine's solr search is down.

yochannah (commented):
Tested again - unfortunately now I can do the following while unauthenticated:

1. Starting from a fresh slate, delete all local storage manually.
2. Do something to prove everything is working, e.g. upload a list,
3. Scramble token (run `bluegenes.events.scrambleTokens()`) to simulate a session expiring
4. Try to do something that requires a token. When the "token has expired" dialogue comes up, clicking on refresh doesn't fix the issue for me - the token-ey action still fails as unauthorised.

heralden:
Sorry about that! I forgot to clear the service token from the dialog button once I added mine-specific auth and persistent anonymous login. Should work fine now.

yochannah:
Note - seems to be working okay when anon and when logged in now. I discovered #399 as part of this but it's not a blocker to merging this PR.

PR #397, opened 2019-08-21
Tool API spec update

### LINKED ISSUE #382: Tool API spec update: don't show tools that have no data

heralden:
PR to resolve https://github.com/intermine/bluegenes/issues/382


heralden:
Note to self: It's probably more correct to use `[:mines mine :service :model :classes]`instead of `[:assets :summary-fields mine]`.

heralden:
I added `depends` to the config.json of the following:

Bluegenes Tissue Expression Visualizer
BlueGenes Tissue Expression Arrayexpress Visualizer
BlueGenes Protein Atlas Tissue Expression Visualizer
BlueGenes Disease Expression ArrayExpress Visualizer
BlueGenes Expression Box Swarm Visualizer

So this PR can be tested by installing the latest version of one of the above, and seeing that the visualisation will only be present on one of flymine or humanmine.

yochannah (approved):

PR #401, opened 2019-08-28
ClojureScript unit tests

heralden:
Adds the kaocha test runner along with some much needed unit tests.

This is a WIP and was created prematurely mostly for the sake of testing CI.

### DIFF
--- a/src/cljs/bluegenes/events/boot.cljs
+++ b/src/cljs/bluegenes/events/boot.cljs
@@ -2,6 +2,7 @@
   (:require [re-frame.core :refer [reg-event-db reg-event-fx subscribe inject-cofx]]
             [bluegenes.db :as db]
             [imcljs.fetch :as fetch]
+            [imcljs.auth :as auth]
             [bluegenes.events.webproperties]
             [bluegenes.events.registry :as registry]
             [bluegenes.route :as route]))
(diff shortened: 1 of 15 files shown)


heralden:
[Kaocha](https://github.com/lambdaisland/kaocha-cljs) has some CI problems (works fine when run locally, but unstable in CI) that should hopefully be fixed in a coming PR in a few weeks.

heralden:
Updating Kaocha has fixed Travis failing it, but now Cypress is suddenly acting up. :hankey:

src/cljs/bluegenes/events/boot.cljs
```diff
@@ -224,43 +225,52 @@
...
-   (if skip?
-     ;; This means tokens have been persisted, so we turn the event handler
-     ;; into a noop just so it gets observed for our boot-flow to progress.
-     db
-     (do (when (nil? token)
-           (.warn js/console "No token available. Nil token will be initialised."))
-         (assoc-in db [:mines (:current-mine db) :service :token] token)))))
+ (fn [db [_ token]]
+   (when (nil? token)
+     (.warn js/console "No token available. Nil token will be initialised."))
```
yochannah, line 272: tiny request - it took me three reads to decide you mean the nil token will be replaced with a real token. Could we change the phrasing? 
Initially I wondered what justification we had for initialising a token as nil. 🤦‍♂

test/cljs/bluegenes/events_test.cljs
```diff
@@ -0,0 +1,160 @@
...
+  (:require [cljs.test :refer-macros [deftest is testing use-fixtures]]
+            [cljs.core.async :refer [chan put!]]
+            [re-frame.core :as rf]
+            [day8.re-frame.test :refer-macros [run-test-sync run-test-async wait-for]]
+            [bluegenes.events]
+            [bluegenes.subs]
+            [imcljs.fetch :as fetch]
+            [imcljs.auth :as auth]))
+
+;; ## Tips on writing re-frame unit tests for the novice
```
yochannah: This is great! Could I suggest we include it in the docs markdown folder instead and just link to it from here though? That makes it easier to find without having to go code spelunking.

src/cljs/bluegenes/events/boot.cljs
heralden: Your first two reads were actually right! It's a warning to let you know that something went wrong and it got a nil token. Not likely to happen, but I have had it occur on older versions of bluegenes, so I thought it would be useful to log it.

test/cljs/bluegenes/events_test.cljs
heralden: That's a good idea. I'll do that now, and probably add more text as well.

src/cljs/bluegenes/events/boot.cljs
yochannah: err, shouldn't we fix it rather than just log it?

heralden: I can't see how it can happen now, really. A better idea would perhaps be to remove the logging.

yochannah: let's just leave it as it is

test/cljs/bluegenes/events_test.cljs
heralden: Done with https://github.com/intermine/bluegenes/pull/401/commits/953816b58a2b64aad99805b4426621476f7fce5d

yochannah (approved):
👍

PR #405, opened 2019-09-27
Fix various bugs when using BlueGenes paired with an InterMine instance

heralden:
- Fix reading serverVars containing InterMine instance data
- Do not show templates for both `:default` and any registry mines previously switched to
- Show informative message when no connection to InterMine instance


yochannah (approved):
We discovered #407 and #406 while trying to test this, but this PR does indeed still work nicely :laughing:

PR #411, opened 2019-09-27
Improve project.clj (particularly lein aliases)

### LINKED ISSUE #406: lein prod doesn't pick up prod config
Steps to reproduce:

1. set your config/dev/config.edn url to something distinctive, e.g. 
```
{ :bluegenes-default-service-root "http://PROD:8080/biotestmine/"
 :bluegenes-default-mine-name "BioTestMine"}
```
2. run lein prod (or `lein cljsbuild once min` and `lein run`)
3. visit the dev cog and look at the URL loaded - it should be from the prod config, but it will in fact be from your dev config.

This is true whether or not the prod config is defined.

### LINKED ISSUE #407: lein dev alias no longer works
Compiling never seems to complete when we run `lein dev` - it sticks partway through compiling the less. If you open the browser, it'll stay on the initial loading screen and show this error in the console: 

image

heralden:
Resolves #406 and resolves #407.

This should make `lein dev` and `lein prod` (and all the other commands I listed on top of *building.md*) fairly robust.

Some things worth noting for future reference:
- `lein-pdo` doesn't work correctly with `less auto` and `figwheel dev` together, so I pushed the invocation of figwheel into `dev/user.clj`
- `lein run` actually starts a dev server (it includes `dev` profile by default)
- `cljsbuild once min` uses the `dev` profile by default, but it usually doesn't matter as it isn't involved with the dev vs. prod configs (the backend handles those)
    - except now that we have `dev/user.clj` starting figwheel, which means you need to use `lein with-profile prod cljsbuild once min` (although it shouldn't be necessary to invoke this by itself, without using the aliases)

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -8,10 +8,22 @@
 * Latest supported [npm](https://www.npmjs.com/)
 * InterMine version 1.8+ (version 2.0 recommended)
 
-
 ## Download NPM dependencies
 
     npm install
+
+## Quickstart
+
+These commands are explained in more depth below, but if you know what you want here's a quick reference of the most useful ones.
+
+    lein dev         # start dev server with hot-reloading
+    lein repl        # start dev server with hot-reloading and nrepl
+    lein prod        # start prod server
+    lein deploy      # build prod release and deploy to clojars
+
+    lein format      # run cljfmt to fix code indentation
+    lein kaocha      # run unit tests
+    npx cypress run  # run cypress ui tests
 
 ## Running a dev environment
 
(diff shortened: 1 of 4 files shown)


docs/building.md
```diff
@@ -8,11 +8,23 @@
 * Latest supported [npm](https://www.npmjs.com/)
 * InterMine version 1.8+ (version 2.0 recommended)
 
-
 ## Download NPM dependencies
 
     npm install
 
+## Quickstart
```
yochannah, line 15: :+1: :+1: :+1: :+1:

yochannah (approved):
tested the new aliases - this is much nicer, good one!

I can also confirm prod config is loaded for prod (and dev is still loaded for dev)

PR #413, opened 2019-10-01
Run biotestmine in CI

heralden:
Testing against Flymine seems to be quite unstable lately, resulting in lots of false positives from Travis. This PR attempts to improve this by starting a biotestmine instance inside Travis CI.

Once this PR is completed, testing is done simply by checking if the automated Travis checks pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
-sudo: required
-jdk: openjdk9
+sudo: true
+jdk: openjdk8
 services: postgresql
 env:
   global:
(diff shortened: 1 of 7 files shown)


yochannah (approved):

yochannah:
approved on the assumption it'll pass - feel free to merge when it does

heralden:
It's sadly not 100% stable, (failed 1/3 runs so far) but seems more stable than using Flymine at least.
Hopefully Daniela's work on slimming down the malaria dataset will improve the stability in the future.

PR #418, opened 2019-10-08
Dynamic page titles

heralden:
Update `document.title` based on the content displayed/page open in the app.

Can be tested by navigating around the app and seeing that the title displayed in the web browser tab changes. Note that there's a bug where switching mines while at the home page won't update the tab title; this will get fixed when #417 is merged in.


yochannah (approved):
tested and works as described - approved to merge if CI passes

PR #419, opened 2019-10-08
Support running bluegenes in https

### LINKED ISSUE #412: Contextually handle HTTP vs HTTPS registry mines
If Bluegenes is running in HTTPS it shouldn't allow you to switch to a non-HTTP mine acquired through the registry. (In fact it probably shouldn't show them at all.)

heralden:
This PR fixes resources not loading due to using http, and filters away insecure http mines when https is active. Fixes #412.

Needs to be used with https://github.com/intermine/imcljs/pull/38. Can be tested by setting up nginx as a HTTPS reverse proxy (any tutorial on the net is applicable). You should see that non-https mines like LocustMine are not present in the mine switcher when bluegenes is run with https.

### DIFF
--- a/src/clj/bluegenes/index.clj
+++ b/src/clj/bluegenes/index.clj
@@ -30,7 +30,7 @@
    css-compiling-style
    [:title "InterMine 2.0 BlueGenes"]
    (include-css "https://cdnjs.cloudflare.com/ajax/libs/gridlex/2.2.0/gridlex.min.css")
-   (include-css "http://cdn.intermine.org/js/intermine/im-tables/2.0.0/main.sandboxed.css")
+   (include-css "https://cdn.intermine.org/js/intermine/im-tables/2.0.0/main.sandboxed.css")
    (include-css "https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css")
    (include-css "/css/site.css")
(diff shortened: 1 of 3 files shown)


heralden:
A simpler way to test this without nginx is to use *run-jetty*'s ssl flag:

1. Generate a keystore with `keytool -genkey -keyalg RSA -alias bluegenes -keystore selfsigned.jks -validity 30 -keysize 2048` and move it into the Bluegenes directory.
1. Add `:ssl? true :ssl-port 4433 :keystore "selfsigned.jks" :key-password "yourpassword"` to the third argument map to `run-jetty` in `src/clj/bluegenes/core.clj`.
1. Start bluegenes and navigate to `https://localhost:4433/` and accept the self-signed certificate.

yochannah (approved):
Tested and works as described on a prod build. Nice one - and thanks for sharing the easy testing tip!

heralden:
@kkarra
I'm mentioning you since you had asked Julie about running Bluegenes in HTTPS a while ago. With this change, there shouldn't be any problem running Bluegenes in HTTPS (preferably served via an Apache or nginx reverse proxy).

We're planning to release this within a few weeks as version 0.9.12, so please be wary that if you're attempting to run an older version of Bluegenes with HTTPS you will likely meet these issues!

kkarra:
thanks for the update, will check it out

PR #420, opened 2019-10-09
Error handling in upload page/ID resolver

### LINKED ISSUE #300: ID Resolver: don't poll the intermine server forever and ever
The main issue is here: https://github.com/intermine/imcljs/issues/19 but there is likely to be a UI component of handling this issue, too.

### LINKED ISSUE #399: upload page should detect expired tokens
on most pages, if you try to do something with an expired token, you'll get an alert that you need to refresh your token. No alert appears on the upload page, but it should. 

A side effect of this is  #300, endless polling of the intermine server when unauthorised.

heralden:
Improve handling HTTP error codes in im-chan effect and redirect to */upload/input* from */upload/save* when no ID resolution in progress (this can happen if you go to the URL directly, or after refreshing an expired token).

Resolves #399 and resolves #300. Dependent on https://github.com/intermine/imcljs/pull/39. Can be tested by running `bluegenes.events.scrambleTokens()` in JS console, then continuing with the example in the upload page.

### DIFF
--- a/src/cljs/bluegenes/components/idresolver/subs.cljs
+++ b/src/cljs/bluegenes/components/idresolver/subs.cljs
@@ -49,6 +49,21 @@
          (fn [db]
            (get-in db [:idresolver :stage :view])))
 
+(reg-sub ::parsing?
+         (fn [db]
+           (= (get-in db [:idresolver :stage :status :action]) :parsing)))
+
+(reg-sub ::parsed?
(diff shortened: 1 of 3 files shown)


yochannah (approved):

yochannah:
Tested locally in prod mode and works nicely. I can also see in he console that the polling doens't go on forever anymore.

PR #422, opened 2019-10-10
Show FASTA information on gene report page

### LINKED ISSUE #45: Gene report page: Add chromosome location and the FASTA sequence to the top of the page.
Add chromosome location and the FASTA sequence to the top of the page. suggested by rachel.

heralden:
This includes chromosome location, FASTA length and sequence. Resolves #45. 

For testing, open any gene (ex. http://localhost:5000/default/report/Gene/1007747) and see that the new information is present.


yochannah (commented):
Technically this works fine, but I suspect some of our labels aren't quite right + I'm not sure if we actually want to display the whole sequence on the screen. 

@rachellyne could you take a look at this and comment? I'll add a screenshot to the PR.

yochannah:
What's on FlyMine: 
https://www.flymine.org/flymine/gene:FBgn0039595
image

Current PR in bluegenes:

image
(the sequence tails offscreen - I suspect we might want to truncate or simply make it downloadable rather than show it)

heralden:
I didn't know FASTA sequences for one gene could be that long. We definitely need to truncate it or add a max height with scroll bars.

I agree that it shouldn't be displayed all the time. I thought we would leave it like this temporarily until we redesign the report page and add the functionality to toggle the displaying of specific blocks, in which case the FASTA sequence would be hidden by default.

yochannah:
I have a feeling that downloading it as a file may be more useful than displaying it at all - @rachellyne will know!

rachellyne:
Yes, most likely users will want to download (or send to a tool).  If it's really long I think fine to just truncate (or scroll bar if that's easy enough).  Sometimes useful to be able to just scan over a sequence - you might want to see which isoform it is maybe, or if it has same start or end to another isoform.

yochannah:
Thanks! @rachellyne - I have one more question for you - are the labels on the screenshot correct? (FASTA sequence, FASTA length)? I thought maybe omitting the word FASTA was better but wanted your input first.

rachellyne:
I would leave FASTA sequence as it is, but for the length I would either put Sequence length or just leave as length if that's a bit too long.

heralden:
I replaced the FASTA viewer with a dropdown button instead and added a download button, as we discussed, and changed the *FASTA length* label to *Sequence length*.

2019-12-06-112412_513x281_scrot

@yochannah Please review again!

yochannah (approved):
@uosl this is great! Downloaded a file and it works as I'd expect. One final teensy change request - add "Download sequence" to the download button, as icons aren't always easy to guess (See "Icons need a text label" here - https://www.nngroup.com/articles/icon-usability/ ) - that should be a minor enough change we can merge as soon as it's done without need for another review, so I'll approve now.

Tests pass locally, both kaocha and cypress.

PR #423, opened 2019-10-11
Pass summary-fields to im-tables-3 to save extra requests

heralden:
Each instance of im-tables-3 has to fetch *summary-fields*, but if we pass it to the table instances, they won't have to do this. This saves one HTTP request per table, which amounts to a lot on the report page where there are close to a hundred im-tables.

This PR can be tested by opening the network panel in your browser's dev tools and observing that a fresh load of a page with one or more im-tables only causes the `summaryfields` endpoint to be requested once (which is when Bluegenes boots).


yochannah (approved):
confirmed - this saves around 100 http requests on a gene report page for flymine! :100:

PR #435, opened 2019-11-04
Use fingerprint compiler option for cache busting

heralden:
Replace appending version number to bundle URL with the ClojureScript fingerprint compiler option. This will give us unique bundle filenames when the code is changed; not just when the version number is bumped. It also allows us to keep a more standard *project.clj* by removing the need for defining the version string outside `defproject`.

Can be tested by running `lein prod` and seeing in the network panel that the bundle `app-XXXXX.js` is downloaded. Then go to the developer page and see that the same hash string is displayed under *client version*. With `lein dev` this should display `app.js` and `dev` instead.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,4 @@
-(def props {:version "0.9.11"})
-
-(defproject org.intermine/bluegenes (:version props)
+(defproject org.intermine/bluegenes "0.9.11"
   :licence "LGPL-2.1-only"
   :description "Bluegenes is a Clojure-powered user interface for InterMine, the biological data warehouse"
   :url "http://www.intermine.org"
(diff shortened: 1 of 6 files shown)


yochannah:
@uosl most of the cypress tests failing on travis pass on my machine - but 

1) Successfully clears invalid anonymous token using dialog 

fails locally on my computer as well. Could you investigate?

heralden:
```
  UI Test
    ✓ Upload page resolved IDs as expected (4089ms)
    ✓ Templates execute and show results (3120ms)
    ✓ Templates allow you to select lists and type in identifiers (3980ms)
    ✓ Gives suggestion results when typing in search (1250ms)
    ✓ Opens the search page to show search results (2137ms)
    ✓ Saves the list from an upload and deletes it (14350ms)
    ✓ Perform a region search using existing example (2074ms)
    ✓ Login and logout works (2286ms)
    ✓ Successfully clears invalid anonymous token using dialog (4584ms)
```
They all passed for me. Are you using `lein prod`? With a dev build, the debug panel will cover the right half of the screen so the dialog won't be clickable.

yochannah:
turns out that sergio had restarted flymine in between when I ran the tests locally and you did 😆 all passes now.

PR #436, opened 2019-11-04
Fix im-tables dropdowns being constrained within div

heralden:
im-tables on the report and list view pages (possibly other places they are used as well) constrain dropdowns within their parent div element. This makes navigating column summaries difficult.

2019-11-04-121415_1191x434_scrot

This PR changes the CSS so that they are displayed overflown past their parent element, as you'd expect dropdowns to look.

We should probably ask ourselves if `overflow-x: auto;` was added to fix something else.

### DIFF
--- a/less/components/report/summary.less
+++ b/less/components/report/summary.less
@@ -10,7 +10,7 @@
         .table {
             margin-bottom: 0;
         }
-        overflow-x: auto;
+        overflow: visible;
     }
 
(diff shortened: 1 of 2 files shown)


yochannah (approved):

yochannah:
one of the cypress tests is timing out, but it's unrelated to this change (sigh).

PR #443, opened 2019-11-20
Support passing POST data to start a region search

heralden:
Pass POST params and body from bluegenes' backend to its SPA, and make it open the region search page and show the results if region data is supplied. This is to support region search queries from external pages.

To test, use an HTML like below, with bluegenes running locally.

```
<html>
  <body>
    <form action="http://localhost:5000" name="list" method="post">
			<input type="hidden" name="features" value="Allele,CDS"/>
			<input type="hidden" name="regions" value="2L:14615455..14619002
2R:5866646..5868384
3R:2578486..2580016"/>
			<input type="submit" value="Submit"/>
		</form>
  </body>
</html>
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -117,7 +117,7 @@
   :profiles {:dev {:dependencies [[binaryage/devtools "0.9.10"]
                                   [day8.re-frame/re-frame-10x "0.4.4"]
                                   [figwheel-sidecar "0.5.19"]
-                                  [cider/piggieback "0.4.1"]]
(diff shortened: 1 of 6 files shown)


yochannah:
@foreveremain - with this endpoint you should be able to pop straight from Jalview into an InterMine - does it look like it would work for you? (we still need the taxon id endpoint for the registry though). 

We can (I think) deploy to a staging server if you want to give it a test in real life.

PR #444, opened 2019-11-27
Mymine folder hierarchy

### LINKED ISSUE #377: Use InterMine List tags to enable folders in bluegenes
tags such as `foldername/subfoldername` should work to enable folders in bluegenes without requiring any further infrastructure. 

Some notes: 
- make sure not to break existing tags in any way. 
- not all intermines will return tags or allow us to set tags correctly. Make sure to test with 1.8.x intermines and see that things still work (in some reasonable way).

heralden:
Depends on https://github.com/intermine/imcljs/pull/42. (Remember to update imcljs dependency before merging this PR.)

Once this is ready, resolves #377.

I will pick this up again once I've received some feedback on how to implement the remaining tasks, and blocking issues are completed.

- [x] Implement save changes with intermine; (see events.cljs:117) **requires** https://github.com/intermine/intermine/issues/2147
- [x] ~~Make sure to test with 1.8.x intermines and see that things still work (in some reasonable way)~~ *Not necessary since we will have to disable this for mines older than when https://github.com/intermine/intermine/issues/2147 is fixed (or first BlueGenes-only release)*
- [x] ~~Record which folders are expanded/collapsed in mymine page into appdb so it will be retained when returning to the page~~ *This is tricky to implement well when the backend doesn't record folders, so we have no way to give consistent unique IDs to folders.*
- [x] Support renaming existing folders in the organize modal (*a rename button beside the folder name could be sufficient?*)
- [x] Support deleting existing folders in the organize modal (*how about creating a "trash node" which you can drag it into, and it will only delete if empty?*) ~~(*folders that have no list children won't get saved, so maybe adding a short message describing this is enough?*)~~
- [x] Show alert when saving changes in organize modal that {{folders}} have no lists in them and therefore won't be saved (*see last note in item above*)
- [x] Go through any remaining TODO comments in code
- [x] Handle error in OrganizeConfirm modal (it should reopen the Organize modal so the error is visible)
- [x] ~~Only enable organize modal (otherwise hide button) if `:enable-unstable-organize` is true in environment~~ The tag manager has been fixed, although we will need to put a version cap (Intermine major version < 5 will have it disabled)
- [x] ~~Touch support *or perhaps hide the button for small screens, since file managers are tough on touch devices*~~ Touch support will be added later with #469

As a potentially useful reference, here's the commit removing folder/directory code from the older mymine infrastructure: https://github.com/intermine/bluegenes/commit/f8f8ee7913113cefc511261dc4231db5a347b82b.


heralden:
This PR is ready for review, although there are a few things that are blocking before we can merge it.

- imcljs PR should get merged and dependency updated
- Discuss the last two items in the TODO list above
- Whether we should wait until we have an InterMine development CI in-place

less/components/organize.less
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    .node-container {
+        display: flex;
+        flex-direction: column;
+        position: relative;
+
+        &.node-fake .node {
+            opacity: 0.65;
+            border: 2px dashed #788896;
```
yochannah: can we refactor the common border styles to be a re-usable mixin? e.g. maybe 
```css
.node-border {
border: 2px dashed #788896;
} 
```

and then include `.node-border;` inside your css rulesets, or alternatively (can't remember if this works)

```css
@node-border: 2px dashed #788896;
 &.node-fake .node {
border: @node-border;
}
```

less/components/organize.less
```diff
@@ -0,0 +1,189 @@
...
+            margin: 0;
+        }
+
+        .icon.symbol-icon {
+            margin-right: 12px;
+        }
+
+        .text {
+            font-size: 1.2em;
+            margin-right: 6px;
```
yochannah: try to use `@spacer` for margins and padding where possible, to keep things uniform / easier to tweak centrally. It's not always practical (borders, 1 or 2px often makes more sense).

less/components/organize.less
```diff
@@ -0,0 +1,189 @@
...
+    .form-control {
+        border-radius: 2px;
+        border-width: 2px;
+    }
+
+    p.error {
+        margin-top: @spacer/2;
+        color: #a94442;
+    }
+}
```
yochannah, line 192: I'm not sure where this popped in, but we have a horizontal scrollbar on all window widths I've tried. This needs to be fixed before merge (happy to help with this if it's being a pain to find! I find CSS relaxing).

src/cljs/bluegenes/pages/mymine/views/modals.cljs
```diff
@@ -327,7 +326,7 @@
                               :type "text"
                               :on-key-up (fn [evt]
                                            (case (oget evt :keyCode)
-                                             13 (do ; Detect "Return
+                                             13 (do ; Detect "Return"
```
yochannah: is there any way to refactor the detect return blocks? I'm guessing it might be impractical for some reason 😆

src/cljs/bluegenes/pages/mymine/views/modals.cljs
```diff
@@ -382,6 +381,61 @@
                                :on-click (fn [] (dispatch [::evts/rename-list name (ocall @input-dom-node :val)]))}
                               "OK"]]]]]])})))
 
+(defn modal-organize []
```
yochannah, line 421: This is _very_ big - can we break it up into 2 or 3 sub components?

less/components/organize.less
heralden: Good idea. I went with the second one (and tested that it works).

yochannah (commented):
Couple more inline comments to be addressed, but the functionality is _great!!_ 

- [x] tried making a folder with a list in it. 
- [x] nested folders
- [x] tried to make a folder with no name - didn't work, gave a good error
- [x] dragged a populated folder into another populated folder. worked as expected. 
- [x] renamed a list
- [x] adding folder works with enter key as well as via mouse
- [x] tried to create and empty folder and got told off in a sensible way.

heralden:
@yochannah I committed fixes for the remaining things you mentioned.

yochannah (approved):
looking great 🚀

PR #446, opened 2019-11-28
Support editing list description

heralden:
Depends on https://github.com/intermine/imcljs/pull/42.

Adds an "add description" button below the list results (and an "edit description" button instead if one is already present).

~~Note that due to a bug in https://github.com/intermine/intermine/issues/2148#issuecomment-559503248, removing an existing description is currently not possible. (Although this will probably be fixed soon.)~~
Update: The above is fixed now.

### DIFF
--- a/src/cljs/bluegenes/pages/results/events.cljs
+++ b/src/cljs/bluegenes/pages/results/events.cljs
@@ -6,6 +6,7 @@
             [imcljs.fetch :as fetch]
             [imcljs.path :as path]
             [imcljs.query :as q]
+            [imcljs.save :as save]
             [bluegenes.interceptors :refer [clear-tooltips]]
             [bluegenes.interceptors :refer [abort-spec]]
             [cljs-time.core :as time]
@@ -163,3 +164,43 @@
                         :enrichment-results
                         (keyword (:widget params))] nil)
       :enrichment/get-enrichment [(:widget params) enrichment-chan]})))
+
+(reg-event-db
+ :description/edit
+ (fn [db [_ state]]
+   (-> db
+       (assoc-in [:results :description :editing?] state)
+       (assoc-in [:results :errors :description] false))))
+
+(reg-event-fx
+ :description/update
+ (fn [{db :db} [_ list-name new-description]]
+   (let [service (get-in db [:mines (:current-mine db) :service])]
+     {:im-chan {:chan (save/im-list-update service list-name {:newDescription new-description})
+                :on-success [:description/update-success]
+                :on-failure [:description/update-failure]}})))
+
+(reg-event-fx
+ :description/update-success
+ (fn [{db :db} [_ {:keys [listDescription listName]}]]
+   {:dispatch [:description/edit false]
+    :db (update-in db [:assets :lists (:current-mine db)]
+                   (fn [lists]
+                     (let [index (first (keep-indexed (fn [i {:keys [name]}]
+                                                        (when (= name listName) i))
+                                                      lists))]
+                       (assoc-in lists [index :description] listDescription))))}))
+
+(reg-event-db
+ :description/update-failure
+ (fn [db [_ {:keys [status body] :as res}]]
+   (assoc-in db [:results :errors :description]
+             (cond
+               (string? body)
+               ;; Server returned HTML as it didn't understand the request.
+               "This feature is not supported in this version of InterMine."
+               (= status 400)
+               "You are not authorised to edit this description."
+               :else
+               (do (.error js/console "Failed imcljs request" res)
+                   "An error occured. Please check your connection and try again.")))))
--- a/src/cljs/bluegenes/pages/results/events.cljs
+++ b/src/cljs/bluegenes/pages/results/events.cljs
@@ -182,14 +182,14 @@
 
 (reg-event-fx
  :description/update-success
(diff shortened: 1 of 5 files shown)


src/cljs/bluegenes/pages/results/events.cljs
```diff
@@ -163,3 +164,43 @@
                         :enrichment-results
                         (keyword (:widget params))] nil)
       :enrichment/get-enrichment [(:widget params) enrichment-chan]})))
+
+(reg-event-db
+ :description/edit
```
yochannah: can I suggest naming this edit-list? otherwise it's a tad vague.

src/cljs/bluegenes/pages/results/events.cljs
```diff
@@ -163,3 +164,43 @@
...
+                       (assoc-in lists [index :description] list-description))))}))
+
+(reg-event-db
+ :description/update-failure
+ (fn [db [_ {:keys [status body] :as res}]]
+   (assoc-in db [:results :errors :description]
+             (cond
+               (string? body)
+               ;; Server returned HTML as it didn't understand the request.
+               "This feature is not supported in this version of InterMine."
```
yochannah, line 201: this works nicely 👍

src/cljs/bluegenes/pages/results/events.cljs
```diff
@@ -163,3 +164,43 @@
...
+
+(reg-event-db
+ :description/edit
+ (fn [db [_ state]]
+   (-> db
+       (assoc-in [:results :description :editing?] state)
+       (assoc-in [:results :errors :description] false))))
+
+(reg-event-fx
+ :description/update
```
yochannah: as above - given this page can be used for non-list queries, all list update/edit functions should probably mention that they're list-specific functions.

src/cljs/bluegenes/pages/results/views.cljs
```diff
@@ -56,11 +56,55 @@
                      [:div.time (time-format/unparse custom-time-formatter (time-coerce/from-long last-executed))]])
                   @historical-queries))])))
 
+(defn description-input [_title initial-text]
```
yochannah: Can we add short docstrings? Thanks!

src/cljs/bluegenes/pages/results/views.cljs
```diff
@@ -72,11 +116,10 @@
              [:div
               {:style {:background-color "white"}}
               [tables/main [:results :table]]
-              (when (> (count description) 0)
-                [:div.description-div
-                 {:style {:background-color "#D2CEBF"  :padding "10px" :overflow "auto"}}
-                 [:b "Description: "]
-                 description])]
+              (when (and list authorized)
+                ;; Only show when results are for a list, not a query.
```
yochannah, line 130: perfect, and I've confirmed this works as expected 💯

yochannah (changes requested):
minor comments added inline - mostly just naming the functions a little more clearly. 

Also - a question - do you think we should add tests for this functionality, either via the cypress UI or through kaocha? I feel like cypress might be easier to implement of the two, but I'm open to other ideas.

src/cljs/bluegenes/pages/results/events.cljs
heralden: What do you think about `:list-description/edit`? This way we can prefix `list` to the other *description* events as well to keep their names consistent.

yochannah: perfect!

heralden:
I agree Cypress would be the most suitable for testing this.

Please check out the changes. If you approve, I'll merge this once https://github.com/intermine/imcljs/pull/42 is merged and updated in bluegenes, and biotestmine is fixed so I can check the Cypress test works.

yochannah:
I've tried testing this with Daniela's bluegenes branch and things aren't working, but I'm not sure  what the culprit is. On the list page, I see a 200 in the network panel for the PUT, but no description appears on the bluegenes list page - this is after an edit:

image

Another thought that's just occurred to me (please don't hate me) - would this make more sense to be in im-tables, perhaps under an edit menu? (edit title, edit contents of list, edit description, etc.)?

I'm also not a massive fan of the brown flat button - I'm guessing you've emulated the previous list pages, but they don't have very strong affordance to suggest things are editable. This might be solved by adding an edit menu to im-tables - maybe chat next week?

heralden:
I tested using biotestmine and a fresh pull of dani's branch, and it worked for me.
Here are all the steps needed in case you forgot something:
1. Build dani's branch on dani's fork of intermine
1. Set imVersion and bioVersion to `5.0.+` in your mine's gradle.properties
1. Start your mine
1. Change the project.clj imcljs version to point to a local install of https://github.com/intermine/imcljs/pull/42
1. Start bluegenes

> would this make more sense to be in im-tables, perhaps under an edit menu?

It might make sense from a UX standpoint, but it would be very weird from a coding standpoint. Remember that the im-table just shows the results of a query; it doesn't know that it's showing a list. 

> I'm also not a massive fan of the brown flat button 

We can always change it later on when we come up with a better design. I think it's most important to have the functionality present.

yochannah (approved):

yochannah:
all works as expected after many trials and travails ✔️  I've successfully edited a list name on alpha flymine.

PR #453, opened 2019-12-03
Fix CircleCI script

heralden:
I heard it's more stable than Travis, making Cypress tests hopefully run more reliably. (It also supports saving test reports and artifacts!)


heralden:
It seems to work fine running the Cypress tests against the Flymine server.
Optimally, we'll later want to create a docker container with an already-built biotestmine instance, and test against that instead.

PR #475, opened 2020-02-12
Better query builder page

### LINKED ISSUE #468: Saving and restoring queries
In the JSP webapp you can save queries and restore them to modify them at a later time in the query builder. We should probably provide this feature in BlueGenes as well.

It seems like the webservices for this are already implemented, as `POST /query/upload` and `GET /query/upload`.

### LINKED ISSUE #470: Support sortOrder/orderBy in query builder

### LINKED ISSUE #471: Support OUTER JOIN in query builder

heralden:
Depends on https://github.com/intermine/imcljs/pull/42.

A makeover with new features of the query builder page.
- Fixes #471
- Fixes #470 
- Fixes #468
- Recent queries
- Data browser
- Import from XML
- And lots of smaller improvements

Following needs to be done before ready to be merged.
- [x] ~~Support creating templates~~ Will do as separate PR
- [x] Test on older InterMine instances


heralden:
@yochannah Can you review? I just realized there's a feature here I need to use in a different PR. 

I will add template creation support in a different PR. I will also test this on older mines myself.

Please let me handle any merge conflicts if they're nontrivial.

less/components/querybuilder.less
```diff
@@ -26,24 +26,183 @@ div.column-container {
...
+            color: #4B5C6B;
+            margin: 0 @spacer/2 @spacer/2 0;
+        }
+
+        .query-table {
+            tbody > tr {
+                cursor: pointer;
+
+                &.active-query {
+                    background-color: #f0fae5 !important;
```
yochannah: !important often opens cans of worms for future maintenance - any chance of eliminating this?

less/components/querybuilder.less
```diff
@@ -278,11 +443,55 @@ div.column-container {
...
+            margin: 0 @spacer/2 0 0;
+        }
+
+        &.active > svg {
+            fill: #039be5;
+        }
+    }
+
+    .constraint-selector {
+        flex-direction: row !important;
```
yochannah: as above - with a more specific selector we _should_ be able to manage without important, I think.

less/components/querybuilder.less
```diff
@@ -543,6 +757,55 @@ div.column-container {
...
+
+            .part {
+                padding: 0 2px;
+            }
+        }
+
+        .sort-button {
+            background-color: white;
+            font-size: 16px;
+            outline: solid 1px #dadada;
```
yochannah: we have `#dadada`, `#fbfbfb` and `#d5d5d5` in a few places - if there aren't important visual distinctions between them, can we refactor to a variable?

less/components/querybuilder.less
```diff
@@ -543,6 +757,55 @@ div.column-container {
...
+        justify-content: space-between;
+        min-height: 40px;
+        max-width: 100%;
+
+        .path-parts {
+            display: flex;
+            flex-wrap: wrap;
+
+            .part {
+                padding: 0 2px;
```
yochannah, line 794: spacer for padding and margin where possible would be nice :)

src/cljs/bluegenes/pages/querybuilder/events.cljs
```diff
@@ -398,11 +444,16 @@
...
               :enhance-query {}
               :order []
+              :sort []
+              :joins #{}
               :preview nil
               :constraint-logic '()
               :im-query nil
               :menu {})))
 
+(defn enhance-constraint-logic [logic]
```
yochannah: could we add a docstring explaining why this is useful?

less/components/querybuilder.less
heralden: I remember struggling with this, failing to create a selector with a higher specificity than the others. I think the usage of !important here is correct since we want the active state to override all the others.

yochannah (changes requested):
Hey @uosl - this is great! Some comments inline, and also... a really long list. Sorry 😭 - I genuinely love the improvements you've already made, but few battle plans survive meeting the yo-enemy. 

-----

on my screen, lein prod produces this overlapping text in the model browser: 
image - same on Chrome and FF.

The sort columns were very low contrast - I can't see the icon very well unless it is clicked on (I have some mild visual impairment and glasses don't fix it perfectly. Others will probably have this complaint too).  
image

In manage columns/query editor, make sure to wrap the buttons and saved query name to the container width:

image
and
image

- If you save a query, I suggest switching to the saved query tab at the bottom. I nearly reported that it failed silently, but then realised I could switch tabs to see it. Alternatively (or maybe as well) a "toast" style notification to let us know it was saved would be good. 
- saved queries when I'm not logged in are transient and will die with the session, right? It would be nice to warn people. Either that or I'm saving them when I'm not logged in and I shouldn't be - not sure which 😆 
- when I click "load" on a saved query, I guess it loads in the query editor window? We should probably auto-focus that tab when we click load. When I have manage columns, or when I'm scrolled down, it's hard to tell if it did anything. Screenshot where I clicked load but see no change: 
image 
- same for the "import xml" load query behaviour - the xml disappears but I don't see any change on the screen when I click load query and have a different tab open or am scrolled down too far. 
- should create template appear when I am not logged in? 
image - oh wait, this template feature is still to be done, isn't it? never mind



----

Longer term wishlist - should I create a ticket? These shouldn't block the merge.
- on the xml tab I wished for a "copy" button.
- failing that, I clicked on the xml and pressed CTRL+A. This selected the whole page - maybe a textarea would be better?

heralden:
I'll need help with this one. Already spent many hours on it before the commit, and I thought adding `height: 100%;` to all `li` elements fixed it. Well it did actually fix it, I have no idea why it reappeared now. Maybe it relied on some cached CSS that's now gone.

less/components/querybuilder.less
heralden: I changed `#dadada` to use `#d5d5d5` and refactored that to a variable. `#fbfbfb` is completely different and only used in one place, so I left that as it is.

less/components/querybuilder.less
heralden: Can we discuss the benefits of using `@spacer` when you're back? There are 40 lines that use margin/padding with px, so it will take me a long time to convert them without breaking alignment (since spacer is 1em, which changes depending on context, meaning I have to use the devtools and my very weary eyes to judge whether it's equivalent, for every single case).

heralden:
> - when I click "load" on a saved query, I guess it loads in the query editor window? We should probably auto-focus that tab when we click load. When I have manage columns, or when I'm scrolled down, it's hard to tell if it did anything.
> - same for the "import xml" load query behaviour - the xml disappears but I don't see any change on the screen when I click load query and have a different tab open or am scrolled down too far. 

Loading a query can change the contents of the model browser, query editor and manage columns, depending on the differences between your active query and the loaded query. I don't think switching to the query editor tab would be the solution, since it's possible that the only thing different with the query you loaded is the sorting or header order (in which case it would only be visible in *Manage Columns*).

I don't know how to make it obvious that you've loaded a query, but I'm open to new ideas. A toast might work, but what text would we put in it? It should say something specific to the query loaded, but recent queries and most query xml don't have a title.

Edit: Recent and saved queries now turn green when they're loaded. Loading an XML query will now show a success message besides the button. Perhaps this is enough as a visual cue?

less/components/querybuilder.less
heralden: ^ This is wrong. I just needed to override the :hover pseudoclass [=

heralden:
I have implemented the remaining things you mentioned, that I haven't commented on.

Don't worry about being picky; many of these things are useful features and necessary fixes!

less/components/querybuilder.less
yochannah: 👍

src/cljs/bluegenes/pages/querybuilder/events.cljs
```diff
@@ -355,6 +370,37 @@
                      :order new-order)
       :dispatch [:qb/enhance-query-build-im-query true]})))
 
+(defn subpath-of? [subpath path]
+  (let [subpaths (->> (split path #"\.")
+                      (iterate drop-last)
+                      (take-while not-empty)
+                      (set))]
+    (contains? subpaths (split subpath #"\."))))
```
yochannah, line 377: not a blocker for merge, but should this be in imcljs or better here? 🌷

heralden: I thought long about the same... Can't remember what tipped me over to leave it out of imcljs.

yochannah (approved):
Okay - please merge once you've looked at my push and are happy to merge 👍 fabulous job!!

PR #482, opened 2020-02-16
Fix Cypress tests

awasthishubh:
### Current Cypress tests have a few issues that are addressed by this PR.

| Before        | After           |
| ------------- |:-------------:|
| Screenshot from 2020-02-15 15-18-56 | Screenshot from 2020-02-16 12-23-46 |

Two test case are still failing which I would like to discuss on discord.

### DIFF
--- a/cypress/integration/pr_test.js
+++ b/cypress/integration/pr_test.js
@@ -54,7 +54,7 @@ describe("UI Test", function() {
         .should("be.gt", 0);
     });
     cy.get("div[class=template-list]").within(() => {
-      cy.get(":nth-child(3) > .col")
+      cy.get(":nth-child(2) > .col")
         .find("button")
         .contains("View >>")
         .click({ force: true });
@@ -63,13 +63,14 @@ describe("UI Test", function() {
     cy.get("@getData").should(xhr => {
       expect(xhr.status).to.equal(200);
     });
-    cy.get("select.constraint-chooser").select("!=");
+    cy.get(".template-constraint-container:nth-child(2) select.constraint-chooser")
+      .select("!=");
   });
 
   it("Gives suggestion results when typing in search", function() {
     cy.get(".home .search").within(() => {
       cy.get("input[class=typeahead-search]").type("mal*");
-      cy.get(".quicksearch-result").should("have.length", 5);
+      cy.get(".quicksearch-result").its('length').should("be.gt", 0);
     });
   });
 
@@ -113,13 +114,13 @@ describe("UI Test", function() {
 
     cy.contains("Add description").click();
     cy.get("textarea").type("My description", { delay: 100 });
-    cy.get("button")
+    cy.get(".controls button")
       .contains("Save")
       .click();
     // Update description to list
     cy.contains("Edit description").click();
     cy.get("textarea").type(" new", { delay: 100 });
-    cy.get("button")
+    cy.get(".controls button")
       .contains("Save")
       .click();
     cy.get(".description").contains("My description new");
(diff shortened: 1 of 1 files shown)


awasthishubh:
Changes: 

> cy.get(":nth-child(3) > .col")

There is an issue with `Adjacent genes (upstream and downstream) + chromosome locations` that I mentioned on Discord. So I changed to second.

#
> cy.get("select.constraint-chooser").select("!=");

This selector selects two elements which throw an error.
#
> cy.get(".quicksearch-result").should("have.length", 5);

To verify Suggestion results, I think a better way will be to check if we have one or more suggestion.
If more genes will be added, this might break.
#
>cy.get("button")
      .contains("Save")

This selector selects two elements. One at the top-right with text `SAVE LIST`
#
> cy.get(".results-body .single-feature")
      .its("length")
      .should("be.gt", 0)

I added this to verify that we have 1 or more results for our query.
#
> cy.get(".results-summary").should("have", "Results");

I don't think this is a proper way to test this. There is no assertion here and if I replace Results with any other text, tests still pass.
#
> cy.get("#password").type("demo");

No ID is set for the password field. Not sure if I should change the selector here or should I set ID in code?
#
>  cy.contains("demo@intermine.org");

In this case, also, there is no assertion. If the element is not found it simply say 0 found. I tried replacing this text and tests passes without fail. I added assertion to check if the matched element is in visible state and it works as expected.

heralden:
Thanks for looking into the Cypress tests! Some things to note, which I sincerely apologize for not being documented:
- Some PRs which depend on new webservices, added via https://github.com/intermine/imcljs/pull/42 have been merged, while imcljs is yet to be updated. This means updating the list description won't work.
- The tests are written for biotestmine. (If you see `.travis.yml`, it uses a script to clone biotestmine and run the Cypress tests against it.)
- Running a local biotestmine can be challenging, (we're working on creating a python CLI tool to automate starting one with docker) especially if it needs the new webservices (eg. updating list description).

I know some of these changes would be improvements when testing against Biotestmine, but it's hard to tell without actually doing it. How about I let you know when the CLI tool is good enough to automate starting a local biotestmine, and then you can look into fixing the Cypress tests again?

Some things I can say for sure:

> No ID is set for the password field. Not sure if I should change the selector here or should I set ID in code?

The ID has been removed, and we'll need to change the selector.

> In this case, also, there is no assertion. If the element is not found it simply say 0 found. I tried replacing this text and tests passes without fail. I added assertion to check if the matched element is in visible state and it works as expected.

It seems like this doesn't do anything, so you can just remove it.

awasthishubh:
@uosl Thanks for the review.
> Some PRs which depend on new web services, added via intermine/imcljs#42 have been merged, while imcljs is yet to be updated. This means updating the list description won't work.

Yes, I do realise that. That's why I left the 6th test as it is. Most of the issues that I tried to fix through this PR, I think, will be there even with the new imcljs.

> How about I let you know when the CLI tool is good enough to automate starting a local biotestmine, and then you can look into fixing the Cypress tests again?

Sure thing! :)

> The ID has been removed, and we'll need to change the selector.

That's what I did.

 > It seems like this doesn't do anything, so you can just remove it.

This verifies that logged in user (visible at the top-right corner)  is the same as the user we log in. I think we should keep this.

heralden (approved):
Sorry for not coming back to you in a long time, but finally the things I mentioned in my previous comment have been resolved.

I looked through your changes properly now and these are really great and well thought out! Thanks for your help. :100:

PR #484, opened 2020-02-20
Handle faulty mines and show loader for slow mines

### LINKED ISSUE #440: Mine name disappears in navbar
When selecting ThaleMine in the mine switcher, it will disappear from the top left of the navbar. (Perhaps because of some missing data from the registry?)

### LINKED ISSUE #474: Avoid showing home page for a glimpse after load before switching page
If you're using a new tab (ie. Bluegenes isn't already open) and open any link that points to a Bluegenes page other than the home page, it will first show the loading screen, then a glimpse of the front page, and then finally the destination page. We should jump straight from the loading screen to the destination page, as showing the home page for half a second makes things seem choppy and could be confusing.

Edit: This could be achieved by moving the `remove wrappy` call to `:finished-loading-assets` (optimally as an effect). We would also need to handle the edge case of an unresponsive mine, perhaps by calling the same effect when showing the mine-loader.

heralden:
Bluegenes isn't very lenient of mines with faulty webservices, or unresponsive servers. This PR teaches Bluegenes to meet other's shortcomings.

An error message is shown if any mine webservice fails, allowing you to continue to use Bluegenes, albeit with a warning of possible reduced capacity. Fixes #440.

All active requests are aborted when switching mines, so you won't get stuck waiting for a timeout when you happened upon an unresponsive mine (you know this has happened as a new mine loader will appear if it takes over 4 seconds). If you wait until the new mine loader disappears (when the browser times out the request), an error will be displayed and page navigation will be disabled, although you will still be able to switch mines.

Error handling for the registry is also added, although it is dependent on #475 to display.

The first-blush/page loader will only be removed once BlueGenes has navigated to the target page. This will cause it to be present slightly longer, but avoid the homepage showing for a glimpse when deep linking. Fixes #474.

All present messages (toasts that appear at the bottom of the page) will now be removed when switching mines. I think this is useful since faulty mines can generate multiple messages that don't disappear by themselves. If the user switches mines, they're probably not interested in them anymore.

You can test:
- an unresponsive default mine by setting `BLUEGENES_DEFAULT_SERVICE_ROOT` to http://8.8.8.9:8080/biotestmine/
- unresponsive and CORS error registry mines by using the diff [1]
- faulty webservices by going through the registry mines until you find one (I promise you will [=)

[1]
```diff
M src/cljs/bluegenes/events/registry.cljs
@@ -54,6 +54,14 @@
                                  (filter #(compatible-protocol? (:url %)))
                                  (map (juxt (comp keyword :namespace) identity)))
                         mines)
+         registry (assoc registry :foomine {:description "Fooey"
+                                            :name "FooMine"
+                                            :url "http://8.8.8.10:8080/biotestmine/"
+                                            :namespace "foomine"})
+         registry (assoc registry :corsmine {:description "Huey"
+                                             :name "CorsMine"
+                                             :url "http://www.flymine.org/flymine"
+                                             :namespace "corsmine"})
          current-mine (:current-mine db)
          db-with-registry (assoc db :registry registry)]
      (cond
```


src/cljs/bluegenes/components/loader.cljs
```diff
@@ -20,3 +21,16 @@
...
+        {:keys [logo name]} @(subscribe [:current-mine])]
+    (when show?
+      [:div.mine-loader
+       [:div.mine-loader-dialog
+        (if (not-empty logo)
+          [:img {:src logo}]
+          [:svg.icon-2x.icon-intermine [:use {:xlinkHref "#icon-intermine"}]])
+        [loader name]
+        [:p "It's taking an unusual amount of time to load this mine."]
+        [:p "Consider contacting the maintainers or switching to a different mine."]]])))
```
yochannah, line 36: would it be a big pain to add the maintainer info from the registry if known?

yochannah: - happy to have that be a ticket for the future if it's easier

heralden: I didn't look into it, since I feel it's very likely that any maintainer email present on the registry is not observed or prepared to handle such enquiries. It's also possible that the problem lies with the user's connection, or the maintainers of the bluegenes instance (ie. not just the mine you're connecting to).
If you still think you want this then you can go ahead and open a ticket. [=

yochannah: seems reasonable!

yochannah (approved):
Okay, ran through the code and tested scenarios 1 and 3 on my machine. All worked as expected. ONE small improvement - we should make some restrictions on max image sizes in that loader. FAWMine has a massive moth logo :) feel free to go ahead and merge this once the image size is fixed.

heralden:
Good catch :rofl:

Issue #465, opened 2020-01-10
"Show me an example" buttons are inconsistent between pages


yochannah:
On the region search [show me an example] is up on the top, in the list upload page there is an EXAMPLE button on the bottom of the page. We should pick one location and be consistent about it, so that users can consistently find it!

itolimaesther:
Hi, I'm an outreachy applicant for May 2020. Please I will love to contribute to this issue.

heralden:
Go ahead, I have assigned you to the issue.

itolimaesther:
Alright Thanks

PR #490, opened 2020-03-10
Show example button is consistent in all pages

itolimaesther:

In other to solve the issue of inconsistent "Example button", I moved the code for the "See example" button from views.cljs file to the bottom right by placing it in results.cljs (before the heading Results)
and I also moved the variable declaration of (ex) to be able to generate the example code.

It addresses "Show me an example" buttons are inconsistent between pages #465

fork and clone the repro the enter the directory
Install Leiningen  if you don't have installed already  => sudo apt install leiningen
npm install
lein dev


heralden:
I feel the placement is still inconsistent. How about placing the *Show Example* button on the region search page, to be directly left of the *Search* button? This should make it pretty similar to the upload page. This would mean that the *Search* button would no longer be full-width.

heralden (changes requested):
Forgot to submit my review. Please see my comment above!

itolimaesther:
Alright @uosl

itolimaesther:
show (1)

@uosl I have corrected it now.

less/site.less
```diff
@@ -416,26 +416,7 @@ button.btn, button.cta {
...
-    position: fixed;
-    top: 0;
-    left: 0;
-    width: 2em;
-    height: 2em;
-    padding: 0;
-    border: none;
-    outline: none;
-    box-shadow: none;
+.btn-default-2{
```
heralden: Looks like something went wrong here during merging? The `p.success`, `p.failure` and `textarea.hidden-clipboard` selectors are needed, so can you add them back?

I also suggest we make the CSS more maintainable by moving the `.btn-default-2` selector into the `less/components/regions.less`, nested appropriately. You should also rename it to `.example-button` so it's more obvious that this styling applies specifically to that one button.

heralden (changes requested):
This is much better :rocket: 

There are some changes to the CSS that needs to be done before I merge this. Hope you can check them out.

less/site.less
itolimaesther: Alright

itolimaesther:
@uosl I have made the changes

heralden (approved):
This looks great. Thank you for your help! :100:

PR #492, opened 2020-03-12
Fix tabs on upload results page not switching

### LINKED ISSUE #479: Unable to change tabs for upload results
On the page `/upload/save`

heralden:
This was due to a dispatch running in the render, causing it to re-run
everytime you switched tabs. The fix was to move it into the on-mount
lifecycle, which required separating the review-step component.

There may be a regression if I messed up shuffling the code about, but I tested all the features I knew of to be present on the page. Fixes #479.


yochannah (approved):
works as expected! 💯

PR #494, opened 2020-03-14
Created 'Useful faq' documentation to help first time contributors

Wyrd00:
It can be confusing trying to orient into the OOS and Intermine community; this document is intended to be a source of information sharing among contributors, many who will invariably run into similar questions and issues and want to share it with the community.


heralden (approved):
Thank you, this will be useful. I think it still needs to be linked from the other documentation, but I'll fix that. [=

PR #511, opened 2020-06-05
Tool improvements and native viz support

heralden:
- Native viz support (see below for details)
  - **Cases** native viz for CovidMine
- Finalise tool support on list/query result page
  - Entity implemented for all classes (it used to be hardcoded to only work for genes)
  - Support for multiple entities to handle multiple query parts
  - Tools and native viz update when editing im-table
- Improvements to tool initialisation
  - Use a re-frame effect instead of side-effects inside `map` in view (which can run multiple times)
  - Remove reliance on subscribe (possible memory leak?)
  - Fix script/link elements being added every time, even when they already exist
  - Avoid Bluegenes halting when a tool throws an error by catching and logging them instead
- Tool API changes
  - Change imEntity to support multiple classes
  - Version key in *config.json*
  - Update documentation and add changelog to Tool API
- Collate version numbers used in Bluegenes into one namespace
  - Better documented and easier to reference

## Rationale for native viz

When creating the CovidMine visualization, it bothered me to no end that if I were to publish it as a Tool, it would probably amount to a 2MB bundle, as well as requiring a lot of boilerplate in a new repo. Since we will include vega-lite in BlueGenes to use in other visualizations, (specifically im-table multi-column summary automated plots) being able to re-use the library for this visualization meant big savings. Therefore I found the need for a *Tool Lite*, which in its simplest form consists of a vega-lite JSON spec along with a pathquery JSON. Since each native viz is just a namespace, code and component re-use is straightforward (something else I found difficult when each tool needed to be its own repository and npm package). It's also possible to re-use the query results across multiple native viz should they happen to use the same query; this will require some small modifications to the event handler running the native viz queries.

Native viz still have a config object similar to *config.json* and they follow the same rules on whether they should be displayed, as tools do. They also appear at the bottom of the tool store page. It is not currently possible to disable them like you can tools, but this can be added later. They follow the same `accepts`, `classes` and `depends` rules so they shouldn't appear when unwanted.

While this is ultimately a product of my laziness, I hope native viz will prove suitable as a lightweight alternative to tools, which just need to run a query and plot the results with vega-lite (which is very feature rich and flexible). If this turns out to be true, modularity and documentation can be added in the future to accomodate outside contributions. Nevertheless, tools are still the *de-facto* way of creating visualizations.

### DIFF
--- a/docs/tools/tool-api.md
+++ b/docs/tools/tool-api.md
@@ -67,13 +67,34 @@ An object representing the data passed to the app, e.g.:
 
 ```json
 {
-  "class": "Gene",
-  "format": "id",
-  "value": 456
+  "Gene": {
+    "class": "Gene",
+    "format": "id",
+    "value": 456
+  }
 }
 ```
 
-To see a full example, look at how the method is called in [bluegenesProtVista's ui demo file](https://github.com/intermine/bluegenesProtVista/blob/master/demo.html) and how [index.js uses the values passed to it](https://github.com/intermine/bluegenesProtVista/blob/master/src/index.js)
+If your tool *accepts* `ids` and takes multiple *classes*, (see [config.json](/docs/tools/tool-api.md#configjson)) it might receive more than one class if they are present on the list or query results page.
+
+```json
+{
+  "Gene": {
+    "class": "Gene",
+    "format": "ids",
+    "value": [1, 2]
+  },
+  "Protein": {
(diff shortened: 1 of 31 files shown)


heralden:
Failing Travis CI is due to the SSL certificate being expired for the bootstrap-material-design docs linked in `developing.md`. Not much we can do with that...

Issue #562, opened 2020-08-25
SVG: Fix icons for difference and subtract set operations


heralden:
The original icons got messed up due to the black-on-white colors being hardcoded into the SVG. This will need some time-consuming manual editing!

heralden:
**To Hacktoberfest participants and other external contributors:**

This requires changing SVG code only, so if you like SVG's or want to learn about them, this is for you!

You can see the icons in the picture below.

![2020-10-02-113602_645x47_scrot](https://user-images.githubusercontent.com/3865590/94914717-712eb980-04a3-11eb-88e5-dbb62e7b68f2.png)

These can be accessed at this page: http://bluegenes.apps.intermine.org/default/lists

Note that they will have a light background by default, but if you select two lists or more (by clicking the checkboxes in the table rows below) the background will turn dark to indicate that they're clickable. The fixed SVG icons should work for both states!

You can use your browser's devtools to copy the code for the SVG icons.

This is how the icons should look like:
- Difference https://en.wikipedia.org/wiki/Symmetric_difference#/media/File:Venn0110.svg
- Subtract https://en.wikipedia.org/wiki/Complement_(set_theory)#/media/File:Venn0010.svg

Once the icons have been fixed, you can use the following tool to convert the HTML to Hiccup: http://htmltohiccup.herokuapp.com/

Finally, to make your PR, copy the Hiccup code and paste it replacing the current Hiccup for the icons:
- Difference https://github.com/intermine/bluegenes/blob/84919744412eaccc30136e80c4f921fac7fbfba2/src/cljs/bluegenes/components/icons.cljs#L105
- Subtract https://github.com/intermine/bluegenes/blob/84919744412eaccc30136e80c4f921fac7fbfba2/src/cljs/bluegenes/components/icons.cljs#L158

Saffafatima12:
Hi! I am new to open source community. I tried solving this issue and I was able to get the design as attached in the images below. Could you please let me know if this solves the problem and if I should proceed to converting HTML to Hiccup?

**1. when two or more in list are selected (dark background)**

![not_selected](https://user-images.githubusercontent.com/52915849/99543801-7e4c2d00-29ab-11eb-8b08-d042e500ad36.PNG)

**2. when less than two selected in list (light background)**

![selected](https://user-images.githubusercontent.com/52915849/99543803-7ee4c380-29ab-11eb-9730-21abc03118b7.PNG)

heralden:
[...]

PR #579, opened 2020-11-18
SVG: Fix icons for difference and subtract set operations #562

Saffafatima12:
Tried fixing the design for icons.

### NEW DESIGN

**1. when two or more in list are selected (dark background)**
svg_click

**2. when less than two selected in list (light background)**
svg_nclick

### OLD DESIGN

**1. when two or more in list are selected (dark background)**
svg_click_old

**2. when less than two selected in list (light background)**
svg_nclick_old

### DIFF
--- a/src/cljs/bluegenes/components/icons.cljs
+++ b/src/cljs/bluegenes/components/icons.cljs
@@ -103,29 +103,29 @@
       {:d
        "M10 14v8h-4v-8h4zM16 6v16h-4v-16h4zM32 24v2h-32v-24h2v22h30zM22 10v12h-4v-12h4zM28 4v18h-4v-18h4z"}]]
     [:svg#icon-venn-disjunction
-  {:viewbox "-5 0 160 100"}
-  [:g#icon-venn-disjunction-g
-   {:transform "translate(1.000000, 1.000000)"}
-   [:circle
-    {:stroke "black",
-     :fill "rgb(170,170,170)",
-     :stroke-width "5",
-     :cy "50",
-     :cx "50",
-     :r "50"}]
-   [:circle
-    {:stroke "black",
-     :fill "rgb(170,170,170)",
-     :stroke-width "5",
-     :cy "50",
-     :cx "100",
-     :r "50"}]
-   [:path#path-diff
-    {:stroke-width "5",
-     :d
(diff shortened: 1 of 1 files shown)


heralden:
The *Combine* icon looks perfect now!

A few things to do first:
1. Change the `:viewbox` property directly below `svg` to be `:viewBox` instead (the "B" needs to be uppercase).
2. Run `lein format` in the bluegenes directory to fix indentation of the new svg source.
3. For *:svg#icon-venn-combine*, change the line below it to have the same viewBox as the other icons: `{:viewBox "-5 0 160 100"}` Without this the scaling will be different. You can also delete the `"\n{:fill-rule \"evenodd\",\n:fill \"none\",\n:stroke-width \"1\",:stroke \"none\"}\n"` line below it, as that was left there before as a mistake and doesn't do anything.

Finally, the new icons look like your screenshot for the dark background, but for the light background they're slightly different:
2020-11-18-170708_1515x120_scrot
You can see that the center part is dark for *Difference* and *Subtract*, when it should be light. It would be great if you're able to fix this.

Thanks again for your help! :+1:

Saffafatima12:
@uosl thank you for your comments! I will need to look into why the the shading in icons turns out to be different from what I see in my browser. I will try my best to correct all the mistakes as soon as possible :)

Saffafatima12:
1. I changed the `:viewbox` to `:viewBox` for all the four icons.
2. Ran lein format in the bluegenes directory to fix indentation of the new svg source.
3. For :svg#icon-venn-combine, I changed the line below it to have the same viewBox as the other icons: {:viewBox "-5 0 160 100"} 
4. Deleted the "\n{:fill-rule \"evenodd\",\n:fill \"none\",\n:stroke-width \"1\",:stroke \"none\"}\n" line.
5. Tried to fix the shading for difference and subtract (Please let me know if it works)

Please do let me know if there are more errors and I will be happy to correct them :) Thank you!

heralden (approved):

heralden:
Thank you for the quick corrections!

Everything looks great now. Thank you so much, this has been a big help :100:

Saffafatima12:
Thank you so much @uosl ! I would have never been able to do this without your help! :)

PR #583, opened 2020-12-11
Smaller UI and UX improvements and fixes

### LINKED ISSUE #98: Search typeahead: better feedback when there are no results
Right now now results == no dropdown. Made me think it was broken at first. (try mashing your fingers on the keyboard to replicate....)

### LINKED ISSUE #577: Handle error/500 response to keyword search
This is particularly important on the search page, where it currently loads indefinitely.

On the navbar and home page quicksearch, it might be useful to show a message as well.

### LINKED ISSUE #578: Button to open template query in query builder
This is present in the JSP webapp, and would be straightforward to add to Bluegenes as well.

heralden:
Various smaller UI and UX improvements. From feedback, usertesting notes and issues (plus deficiencies I discovered while fixing these; *I have bad self-control*).

Each problem is noted in the commit text.

One large change involves removing the report page layout *default* configuration. It is no longer possible to define a *default* layout, for use instead of the fallback when a *class-specific* layout isn't defined.
This was removed as it required utilizing the model property's `referencedType` for the items in the layout. This led to a bug where classes which used one `referencedType` for multiple items (e.g. Protein which has a *Canonical Protein* reference in addition to an *Isoforms* collection, both which have a `referencedType` of *Protein*), meant the report page wasn't able to distinguish between which to display. Layouts now use the model property's `name`, which fixes this bug but makes the existing *default* configuration not applicable.
It doesn't seem like the *default* layout could achieve its original vision, and the fallback layout seems more suitable, so I don't think this is a loss.

Issues not linked from commits: Fixes #578, fixes #577, fixes #98


heralden:
CI is failing because current Bluegenes depends on a different Intermine branch.

Tests pass locally against biotestmine.

PR #588, opened 2020-12-17
Delete package-lock.json

sammyjava:
This shouldn't be in the GitHub repo since it gets recreated by npm install, and is slightly different each time.

What I did:
- deleted package-lock.json
- prevents a modified file from being created in an untouched repo clone
- no testing required, one always has to run npm install


heralden:
The package-lock.json file is the only thing that ensures reproducible builds when running `npm install` (otherwise npm will install the latest minor version for all packages depended on with a caret ^x.x.x, which is default). When it gets generated, npm also logs "created a lockfile as package-lock.json. You should commit this file." (More info here: https://stackoverflow.com/a/44210813)

It's a source of confusion though, as most package managers don't aggressively adhere to semver, removing the  need for a lockfile.

Does this clear it up, or did I misunderstand your argument?

sammyjava:
No, I didn't know that! I thought it was just for caching purposes after you've done an npm install.

PR #617, opened 2021-02-02
Support for login via OAuth2 providers and improve backend

### LINKED ISSUE #445: Avoid sending model to bluegenes backend
We often pass `service` to the bluegenes API in requests from the client.

Model is often included in this map, even if it's not used, making the payload much much larger than it needs to be.

### LINKED ISSUE #519: Support OAuth2 login
~~This is to allow users to have an account across all intermines accessible from Bluegenes, without having to create an account on each one and later login to each one separately.~~ This is only partly correct; you only need to have one account, but you'll need to run through the OAuth2 process for each mine, and each mine will need to have the OAuth2 provider configured.

This requires work on the InterMine backend first: https://github.com/intermine/intermine/issues/1720

### LINKED ISSUE #521: Notify with message of renamed conflicting lists on login
This occurs when you create a list in an anonymous session, with the same name as a list under an account, then login to that account such that the backend has to rename your new list to save it under the account.

This will require more work than initially expected, as the data on renamed lists is present in the login response, and login/logout is handled on Bluegenes' backend (so it can handle sessions, although we could consider removing this as we currently depend on localstorage for sessions) ~~with imcljs expecting the login response to only return a token~~ this is wrong; it's Bluegenes' backend which extracts the token.

heralden:
Add buttons for logging in using an account on a supported OAuth2 provider (currently Google and Elixir AAI). These need to be configured on the Intermine instance (see https://github.com/intermine/intermine/issues/1720).

Also cleans up Bluegenes auth API and notifies the user when lists from an anonymous session are automatically renamed due to merge conflicts with existing lists when logging in.

Finally, fixes some issues found with the Bluegenes backend:
- Due to charset mismatch between HTML and Content-Type header, Firefox' initial request would be done twice (this also caused `session.init` to not work for Firefox, which is used for permanent URL resolution errors and now OAuth2 login)
- All HTTP response Content-Type was application/json instead of the correct mimetype for each resource
- Stricter cookie attributes and recreate when logging in (privilege escalation)

Fixes #519
Fixes #521
Fixes #445


heralden:
Failed tests are due to being unable to build Intermine bluegenes branch.

PR #696, opened 2021-03-29
fix: discrepancies in lock file after setting up project

### LINKED ISSUE #695: Discrepancy in package-lock.json after setting up project
**Current Behaviour**
After doing `npm install`, there are several discrepancy in the package-lock.json file

**Actual Behaviour**
No discrepancy in package versions should be present

**OS**
Ubuntu

meenal06:
fixes #695 

- After setting up project through `npm install` there are various changes in lockfile done.
-#965
- Run `npm install` after checking out the branch

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,35 @@
 {
   "name": "bluegenes",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "@cljs-oss/module-deps": {
+  "packages": {
+    "": {
+      "version": "1.0.0",
+      "license": "SEE LICENSE IN LICENCE",
+      "dependencies": {
+        "@cljs-oss/module-deps": "^1.1.1",
+        "bootstrap": "^3.3.7",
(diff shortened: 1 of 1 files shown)


meenal06:
Hi @uosl , looks like the failure is not due to my PR but cypress complaining

heralden:
> Hi @uosl , looks like the failure is not due to my PR but cypress complaining

You're right. Some of the current Cypress tests can be unstable (hoping we can improve this), so some will fail at random.

heralden:
What version of node.js and npm are you using? I'm using v14.16.0 and 6.4.11, and when running `npm install` on your branch it will revert to the old lock file. It looks like this package lock file is of a newer version. I will have to investigate whether they are backwards compatible and they recommend upgrading.

meenal06:
I am on node v15.7 and nom v7.4.3, also the updates are backwards compatible. I have verified building the project with the new changes.

heralden:
A problem is that when running `npm install` with an older version, it will rewrite the lockfile into the older format. I assume that most people will be using v14 which is on Active LTS (odd-numbered releases are considered unstable by Node). I would rather wait until node v16 enters Active LTS and becomes more widespread, before updating the lockfile.

Thank you for letting us know about this. I'll be closing the PR.

PR #697, opened 2021-03-30
fix: remove support , fixtures and plugins from gitignore

meenal06:
- I have restored cypress support, fixtures folder
- I have not created an issue for this
- No explicit tests needed, the following folders are core to cypress as they have all got different usages. I don't think we should ignore them as they are helpers for cypress main tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,10 +8,7 @@
 /resources/public/css
 /resources/public/tools/*
 /cypress/screenshots
-/cypress/plugins
 /cypress/videos
-/cypress/support
-/cypress/fixtures
 /cypress/results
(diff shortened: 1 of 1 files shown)


meenal06:
Hi @uosl , The cypress support and fixtures folders are important for cypress, as I can define any alias command or any fixture to be used under the tests. They don't have any use in current tests written that's why they might have been places under gitignore but they are helpers to cypress core which is why I have restored them and will commit their content in a successive PR.

heralden (approved):
Thanks for spotting this! Looks like they've been wrongfully added to gitignore.

Issue #700, opened 2021-03-31
Tests: Reorganise specs, remove flakes


meenal06:
Cypress tests needs to be reorganised a bit in terms of spec naming, file naming and spec placement as well as use of before-each in the tests.
Also there are flakes in cypress right now which needs to be addressed

PR #701, opened 2021-03-31
feat: reorganise and fix login specs

meenal06:

- reorganised the login spec
- refers #700
- checkout on local and run with cypress


meenal06:
Hi @uosl , This PR fixes and reorganises the login specs. If this looks good to you, I will make a successive PR for registration specs too.

heralden:
The diff looks great, this is much cleaner :+1:. However, it fails with `TypeError: cy.openLoginDialogue`. I assume this is because this function is defined in the folders that used to be in gitignore until #697. Can you please add it to this PR before I merge?

meenal06:
Hi @uosl , I have updated the PR please have a look!

heralden (approved):

heralden:
Judging by the artifacts on CircleCI, the single login-logout test that failed seems to be a flake. We'll improve on this later.

Thanks for your contribution!

PR #707, opened 2021-04-09
feat: improvise register_spec tests

meenal06:
Signed-off-by: Meenal Trivedi <meenaltrivedi6102@gmail.com>

- Fixes the flakes and errors in the register_spec, makes sure that the ui tests are attached to ui perspective only.
- checkout on local and run cypress


meenal06:
Hi @uosl Please review the following pr :smiley: !

heralden (approved):
Looks good! Good idea to use a random email, so it won't fail on repeat runs. :+1:

PR #731, opened 2021-06-28
Added more cypress tests

jiaq8877:


cypress/support/commands.js
```diff
@@ -33,3 +35,48 @@ Cypress.Commands.add("openRegisterDialogue", () => {
...
+})
+
+Cypress.Commands.add("openQueryBuilderTab", () => {
+    cy.get("#bluegenes-main-nav").within(() => {
+        cy.contains("Query Builder").click();
+    });
+})
+
+Cypress.Commands.add("searchKeyword", (keyword) => {
+    cy.get(".search").first().type(keyword + '{enter}',{delay:100});
```
heralden: For `Search Test can search two terms with OR`, it failed 3 times to type in *CDPK1 OR CDPK4* correctly. Instead of increasing the delay, what do you think about using `cy.visit("/biotestmine/search");` in the beforeEach for these tests, and then using the search input in the page instead of the one in the navbar? This one won't return suggested results, so it should be less likely to fail in CI.

cypress/support/commands.js
```diff
@@ -33,3 +35,48 @@ Cypress.Commands.add("openRegisterDialogue", () => {
...
+})
+
+Cypress.Commands.add("createList", () => {
```
heralden: Does this work? I can't see it used anywhere so I'm not sure. Those commands inside `cy.get("#bluegenes-main-nav").within` seem to apply to elsewhere than the navbar. It would be better to use `cy.get(".wizard").within`. The selector for adding the identifiers could possibly be simplified to `cy.get('textarea')`.

cypress/integration/upload_test.js
```diff
@@ -0,0 +1,38 @@
...
+        cy.get('.identifier-input > .form-group > .form-control').type("ABRA, CRK2, CDPK1, CDPK4",{delay:100});
+        cy.contains("Continue").click();
+        cy.url().should("include","/save");
+        cy.contains("Save List").click();
+        cy.url().should("include","/results");
+
+        cy.openListsTab();
+        cy.url().should("include","/lists");
+        cy.get(".lists-item").its('length').should("be.gt", 0);
+    });
```
heralden: Similarly to my comment on the `createList` command, have you tried putting inside `within .wizard` and using `textarea` for the identifiers input?

cypress/integration/templates_test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it("can be accessed from the main page", function() {
+        cy.contains("More queries here").click();
+        cy.url().should("include", "/templates");
+    });
+
+    it("can filter templates by search", function() {
+        cy.openTemplatesTab();
+        cy.url().should("include", "/templates");
+        cy.get('input[class="form-control input-lg"]').type("Organism{enter}",{delay:100});
```
heralden: Did you try selecting this based on the *Filter by description* label? They're siblings so you might be able to do `cy.contains('Filter by description').next()`.

cypress/integration/templates_test.js
```diff
@@ -0,0 +1,45 @@
...
+        cy.get('#Organism_Protein').should("have.id","Organism_Protein");
```
heralden: I think it would be better to test whether the first `.template` element has ID `Organism_Protein`. I think this assertion would still succeed even if the filter didn't do anything.

cypress/integration/templates_test.js
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    it("can search templates by categories", function() {
+        cy.openTemplatesTab();
+        cy.url().should("include","/templates");
+        cy.get("div[class=grid-1]").should("have.length","3");
+        cy.contains("Genomics").click();
+        cy.get("div[class=grid-1]").should("have.length","1");
```
heralden: `div[class=grid-1]` is more likely to change than `.template`. Have you tried selecting for `.template` and checking length of that?

cypress/integration/search_test.js
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    it("can search for partial matches", function(){
+        cy.searchKeyword('MAL*');        
+        cy.url().should("include","/search");
+
+        cy.get(".results > form").children().first().click();
+        cy.url().should("include","/report");
+        cy.get('.report-table-body').within(() => {
+            cy.get(".report-table-cell").filter(':contains("MAL")');
```
heralden: Are you sure this will fail if there's no MAL? Maybe you have to add `.its('length').should("be.gt", 0);`

cypress/integration/query_builder_test.js
```diff
@@ -0,0 +1,190 @@
+describe("Query builder test", function() {
+    beforeEach(function() {
+      cy.visit("/biotestmine/querybuilder");
+    });
+
+    it("allows you to select data type and summary attributes with model browser", function() {
+      cy.get('.model-browser-root').within(() => {
+        cy.get('select').then($option => {$option.val("Protein")}).parent().trigger('change');
```
heralden: Can you add a cypress command for this (taking the class you want to select as argument)? It's used many places and has complex syntax.

cypress/integration/query_builder_test.js
```diff
@@ -0,0 +1,190 @@
...
+    it('can add filters on query editor', function(){
+      cy.get('.model-browser-root').within(() => {
+        cy.get('select').then($option => {$option.val('Gene')}).parent().trigger('change');
+        cy.get('select').should('have.value', 'Gene');
+        cy.contains("Summary").click();
+      });
+
+      cy.get('.query-view-column').within(() => {
+        cy.get('.qb-label').should('include.text','Gene');
+        cy.get(':nth-child(1) > :nth-child(1) > :nth-child(1) > .lab > .icon-filter').click(); //rewrite
```
heralden: Have you tried: `cy.contains('Organism').parents('.qb-label').siblings('.icon-filter').click()` where *Organism* is the attribute/class name?
Not much shorter, but at least the intent is clearer.

heralden: I see now that you arrived at something similar a few lines below :smile:

cypress/integration/query_builder_test.js
```diff
@@ -0,0 +1,190 @@
...
+        
+        cy.contains("Saved Queries").click();
+        cy.get('table').contains('td','Protein summary'); //Assertion
+
+        cy.contains('Rename').click();
+        cy.get('input.form-control').clear();
+        cy.get('input.form-control').click().type('Gene summary{enter}',{delay:100});
+        // cy.contains(/^Save$/).click();
+        // cy.get('[title="Load this query"]').find('.td').should('have.text','Gene summary');
+        cy.get('.panel-body > .table > tbody > tr > td').first().should('include.text','Gene summary');
```
heralden: How about: `cy.get('.query-table').contains('td', 'Gene summary');`

PR #734, opened 2021-07-05
Cypress lists and report pages

jiaq8877:


cypress/integration/report_page_test.js
```diff
@@ -0,0 +1,104 @@
...
+        cy.get('.report-table-heading').contains("Visualizations")
+        .parent(".report-table")
+        .as('visualizationsSection');
+
+        cy.get("@visualizationsSection").within(() => {
+            cy.get(".report-item").should("exist");
+            cy.get(".icon-chevron-up").click();
+            cy.get(".report-item").should("not.exist");
+        })
+    });
```
heralden: Can you make this use the *Data* heading instead? The visualizations may change in the future, and the header will disappear if there aren't any.

cypress/integration/report_page_test.js
```diff
@@ -0,0 +1,104 @@
...
+        cy.get('.report-table-heading').contains("Visualizations")
+        .parent(".report-table").find(".report-item").filter(':contains("MSA Viewer")')
+        .as('MSAViewerSection');
+
+        cy.get("@MSAViewerSection").within(() => {
+            cy.get('.bluegenesMSAViewer').should("not.be.visible");
+            cy.get('.report-item-toggle').click();
+            cy.get('.bluegenesMSAViewer').should("be.visible");
+        })
+    });
```
heralden: Similarly to my last comment, can you change this test to run on a report item under *Data*, that contains an im-table instead of visualization? It failed in CI because that biotestmine was missing the MSA viewer (it can happen that it doesn't get time to install if the tests run too fast).

cypress/integration/report_page_test.js
```diff
@@ -0,0 +1,104 @@
+describe("Report Page Test", function(){
+    beforeEach(function(){
+        cy.searchKeyword("GST");
+        cy.get(".result").click();
+        cy.url().should("include","/report");
+    });
+
+    it("can filter the topic of a search result", function(){
+        cy.get("main").find("input").type("Publications{enter}",{delay:100});
```
heralden: This failed 3 times in CI and the screenshot shows it failed before the report page finished loading.
Can you try this instead: `cy.get(".report-page-filter").find("input").type("Publications{enter}",{delay:100});`

cypress/support/commands.js
```diff
@@ -43,13 +43,57 @@ Cypress.Commands.add("openListsTab", () => {
...
 Cypress.Commands.add("selectFromDropdown", (keyword) => {
     cy.get('select').then($option => {$option.val(keyword)}).parent().trigger('change');
 })
 
+Cypress.Commands.add("createGeneList", (geneList) => {
+    cy.contains("Upload").click();
+    cy.get(".wizard").find("textarea").type(geneList,{delay:100});
+    cy.contains("Continue").click();
+    cy.contains("Save List").click();
+    cy.wait(100);
```
heralden: When I look at the CI test recording, it seems this wait isn't always enough. What will happen is that it will click the *Save List* button, then click *Lists* and Bluegenes will change to the lists page first and then to the query results page (as it waits for the list to be saved successfully before showing it, which makes it occur after changing to the list page).

Can you try replacing the `wait` with `cy.url().should("include","/results");`? The assertion should force cy.url() to retry until it passes, meaning Cypress won't continue until the page has actually changed.

cypress/integration/report_page_test.js
```diff
@@ -0,0 +1,104 @@
+describe("Report Page Test", function(){
+    beforeEach(function(){
+        cy.searchKeyword("GST");
+        cy.get(".result").click();
+        cy.url().should("include","/report");
```
heralden, line 5: The URL will change before the contents change, so I think a better assertion to block Cypress from continuing until the page is ready, is to query for a DOM element that only exists when the page is ready:
`cy.get('.report-page-heading').should("exist");`
Some useful info about this: https://sqa.stackexchange.com/questions/44539/how-to-wait-till-all-components-are-loaded-in-cypress

cypress/integration/lists_test/single_list_test.js
```diff
@@ -0,0 +1,71 @@
+describe("Single Lists Test", function() {
+    beforeEach(function() {
+      cy.visit("/");
+      cy.createGeneList("SODB, GBP, GST, CDPK1");
+    });
+
+    it("can copy a list", function(){
+        cy.url().should("include","/lists");
+        cy.get('.lists-item').should('have.length',1);
```
heralden: Any reason why these lines can't be in the beforeEach, similar to multiple_lists_test?

jiaq8877: Nope, just added them to beforeEach!

cypress/integration/report_page_test.js
```diff
@@ -0,0 +1,105 @@
...
+                cy.get("input.form-control").should("be.visible").invoke('val').should("include","http://localhost:9999/biotestmine/gene:PF14_0187");
+                cy.get("button").contains("Copy").should("be.visible").click({multiple:true});
+            });
+        })
+        // cy.wait(500);
+        // cy.task("getClipboard").should("contain","/biotestmine/gene:PF14_0187");
+    });
+
+    it("can check if the gene is present in any list", function(){
+        cy.get(".col-lg-12").filter(':contains("Lists")').should("be.visible").as('listsSection');
```
heralden: Can you change `.col-lg-12` to `.sidebar-entry` if it works? The col classes are for layout and may change.

cypress/integration/report_page_test.js
```diff
@@ -0,0 +1,105 @@
...
+        cy.get("@exonsSection").within(() => {
+            cy.get('.im-table').should("exist");
+            cy.get('.report-item-toggle').click();
+            cy.get('.im-table').should("not.exist");
+        })
+    });
+
+    
+    it("can copy the permanent URL of the report page of the object", function(){
+        cy.get(".visible-md-block").within(() => {
```
heralden: This `.visible-md-block` class may change as well.
`cy.get(".sidebar-actions").filter(':visible').within(() => {` might work.

PR #742, opened 2021-07-22
Cypress tests

jiaq8877:
- Added tests on account-specific functions
- Added and reorganised query builder tests
- Flaky tests: admin_account_test, report_page_layout_test, account_list_test
- (Somewhat) Flaky tests: search_test, upload_test


cypress/integration/query_builder_test/query_view_test.js
```diff
@@ -0,0 +1,148 @@
...
+
+    it("can remove attributes on query editor", function(){
+      cy.get(".model-browser-root").within(() => {
+        cy.selectFromDropdown("Protein");        
+        cy.get("select").should("have.value", "Protein");
+        cy.contains("Summary").click();
+      });
+
+      cy.get('.query-view-column').within(() => {
+        cy.contains('Protein').parentsUntil('.lab qb-class').children('.icon.icon-bin').click();
```
heralden, line 52: Can you add an assertion after this, to verify it's been removed?

cypress/integration/lists_test/account_list_test.js
```diff
@@ -0,0 +1,158 @@
...
+            cy.get('[type="checkbox"]').check();
+        })
+        cy.get('.bottom-controls').within(() => {
+            cy.contains("Copy all").click();
+        })
+        cy.get(".css-1wy0on6").click().type("List 2{enter}",{delay:100},{force:true});
+        cy.get(".modal-footer").within(() => {
+            cy.contains("Copy list(s)").click();
+        })
+    })
```
heralden: Can you add assertions to the end of these two tests, checking that the folder with that name is visible?

cypress/integration/admin_account_test/report_page_layout_test.js
```diff
@@ -0,0 +1,100 @@
...
+		cy.get("input#email").type("test_user@mail_account");
+		cy.get("input[type='password']").type("secret");
+        cy.intercept('POST', '/api/auth/login').as('login');
+		cy.get(".login-form")
+			.find("button")
+			.contains('Login')
+		 	.click();
+        cy.wait('@login');
+        cy.get(".logon.dropdown.success").should("exist").click();
+        cy.get(".logon.dropdown.success").should("contain", "test_user@mail_account"); //flaky
```
heralden: What do you think about adding a cypress command for logging in as admin? I've seen this snippet used several places.

cypress/integration/admin_account_test/report_page_layout_test.js
```diff
@@ -0,0 +1,100 @@
...
+                cy.get("textarea").clear().type("Coding regions within this gene{enter}",{delay:100});
+                cy.contains("Save").click();
+            })
+
+            cy.contains("Save changes").click();
+            cy.get(".success").should("include.text","Successfully saved changes to report page layout.");  
+        })
+    })
+
+    it("can edit the visibility of the description of a class", function(){
```
heralden: This test name is misleading - the "eye" button is used to collapse the content by default (table or visualization on the report page) so that the user has to click to expand if they wish to view it. Can you change it to "can make a class show as collapsed"?

cypress/integration/admin_account_test/report_page_layout_test.js
```diff
@@ -0,0 +1,100 @@
...
+            cy.get("ul").filter(".classes").find(".btn-fab").last().click();
+            cy.get(".btn-group-sm").eq(1).children().eq(4).click({delay:100});
+
+            cy.contains("Save changes").click();
+            cy.get(".success").should("include.text","Successfully saved changes to report page layout.");  
+        })
+    })
+
+    after(function(){
+        cy.visit("/biotestmine/report/MRNA/2000009");
```
heralden: This hardcoded ID can become invalid in the future. I suggest you use the URL generated in *Copy permanent URL* from its report page instead.

Also, can you move this into a test instead of having it in the after hook? I think those won't be retried, while it seems tests are run in order within a suite, so it should be fine to have it as a regular test.

cypress/integration/admin_account_test/report_page_layout_test.js
```diff
@@ -0,0 +1,100 @@
...
+        cy.get(".report-page-heading").should("include.text","mRNA.46312");
+        cy.get('.report-table-heading').contains("Interactions")
+        .parent(".report-table")
+        .as('interactionsSection');
+
+        cy.get("@interactionsSection").within(() => {
+            cy.get(".report-item-title").should("include.text","Exons");
+            cy.get('[class="icon icon-info"]').should("exist").click();
+            cy.get(".report-item-title > .dropdown > .dropdown-menu").should("exist")
+            .find("p").should("include.text","Coding regions within this gene");
```
heralden, line 150: Can you also assert that the Exons' `.im-table` isn't present (due to clicking the "eye" button as I explained above), then click the Exons header and assert that `.im-table` is now present?

heralden (approved):

PR #747, opened 2021-08-03
Cypress regions tests

jiaq8877:


cypress/integration/user_account/user_account_test.js
```diff
@@ -0,0 +1,146 @@
...
+        cy.createProteinList("Q8ID23_PLAF7, Q6LFN1");
+        cy.get('.lists-item').its("length").should('be.gt',0);
+
+        cy.contains("Tags").parent().find(".icon-selection").click();
+        cy.get(".dropdown").find("li").filter(':contains("Algae genes")').click();
+        cy.get('.lists-item').should('have.length',1);
+        cy.get('.lists-item').find(".list-title").should("include.text","List of common genes");
+    })
+
+    it("can build and save a query", function(){
```
heralden: After saving the query, can you add another `cy.visit("/biotestmine/querybuilder");` (to make sure the session gets cleared - you might have to login again) and then assert that the saved query is still there?

cypress/integration/regions_test/regions_export_test.js
```diff
@@ -0,0 +1,93 @@
...
+            cy.wait("@regionSearch");
+        })
+    });
+
+    it("can export results of region search in TAB", function(){
+        cy.window().document().then(function (doc) {
+            doc.addEventListener('click', () => {
+              setTimeout(function () { doc.location.reload() }, 5000)
+            })
+        })
```
heralden, line 18: Did you try moving this into `beforeEach`? Would also be a good idea to add a comment with the link to where this is from (github issue I think), as well as comment that it's a workaround to avoid a page timeout failure.

cypress/integration/regions_test/regions_export_test.js
```diff
@@ -0,0 +1,93 @@
+describe("Regions Export Test", function(){
```
heralden, line 1: Could you repeat these tests for the first region result, using the buttons inside `.results-footer`? They give different files (if you have more than one region results - let me know if this isn't the case for the biotestmine example) and could fail while the exports above for all results continue to work.

In this case it would be useful to have more descriptive filenames for the fixtures, like `all_regions_result.csv` and `region_result.csv`.

cypress/integration/regions_test/regions_test.js
```diff
@@ -0,0 +1,177 @@
...
+                })
+            })
+            cy.get("button").filter(':contains("Search")').click();
+        })
+        cy.get(".results").within(() => {
+            cy.get("#region-result-0 > span").should("include.text","MAL1 1..30000");
+        })
+    })
+
+    it("can select to perform strand-specific region search", function(){
```
heralden: Can you extend this test to also search the reverse strand, meaning
```
MAL1:0..30000
MAL1:30000..0
```
and checking that the second result includes arrow-left.

cypress/integration/regions_test/regions_test.js
```diff
@@ -0,0 +1,177 @@
...
+        cy.get(".extend-region").within(() => {
+            cy.get(".icon-lock").click();
+            cy.get(".slider-ticks").eq(0).within(() => {
+                cy.contains("100k").click();
+                cy.wait(500);
+            })
+            cy.get("input.form-control").eq(1).should("have.value","100k");
+            cy.get("input.form-control").eq(2).should("have.value","0");
+        })
+    })
```
heralden: Can you in addition, run a region search and check that the header for the result is adjusted correctly, e.g. `MAL1:0..30000` becomes `0..130000` for it "can extend gene search region by clicking tick marks."

Beware of overly large regions that could cause the tests to take a long time to complete!
It would also be great if you can use a variety of values for extend regions, like numbers only, numbers with k, numbers with decimal and k. Using 0 as start region for one test is fine, as that tests it shouldn't go negative, but add some searches that don't start at 0 so you can test that the starting location gets extended as well.

heralden (approved):

heralden:
Test that failed in CI `"Opens the search page to show search results"` is an old test that's been fixed in dev branch.

PR #750, opened 2021-08-12
Cypress enrichment tests

jiaq8877:


cypress/integration/lists_test/enrichment_test.js
```diff
@@ -0,0 +1,99 @@
...
+})
+
+describe("Enrichment Test", function(){
+    beforeEach(function(){
+        cy.visit("/");
+        cy.createListFromFile("all_genes.csv","All genes"); //This file is downloaded from querybuilder.
+        cy.createListFromFile("enrichment_gene.csv","Enrichment analysis genes");
+        cy.get(".enrichment").should("exist").within(() => {
+            cy.get(".dropdown").click();
+            cy.get(".list-selection").eq(0).click();
```
heralden: It appears you are setting the background population for every test. Can you instead only do this for one test, which is to specifically test the background population? When it isn't set, the enrichment will use all the genes in the mine as background population.

cypress/integration/lists_test/enrichment_test.js
```diff
@@ -0,0 +1,99 @@
...
+            cy.contains("View").click();
+            cy.wait("@tableLoad");
+        })
+        cy.get('.pagination-label').should("include.text","172 rows");
+        cy.get(".enrichment-category").eq(0).within(() => { 
+            cy.get(".icon-download").click();
+        })
+        cy.readFile("cypress/downloads/Enrichment Results Gene Ontology Enrichment.tsv").then(newResult => {
+            cy.readFile("cypress/fixtures/all_enrichment_result.tsv").should("eq",newResult);
+        })
```
heralden: In this and the other "view and download" tests, can you switch the order so it downloads first, and then *views*?

cypress/integration/lists_test/enrichment_test.js
```diff
@@ -0,0 +1,99 @@
...
+            cy.intercept("POST","/biotestmine/service/list/enrichment").as("enrichmentLoad");
+            cy.contains("Max p-value").parent().find("select").eq(0).select("0.10");
+            cy.get(".correction").find("select").select("Benjamini Hochberg");
+            cy.get(".text-filter > .form-control").type("x");
+            cy.get(".text-filter > .form-control").click().wait(500).clear().type("symbiosis{enter}",{delay:100}); //very flaky
+            cy.wait("@enrichmentLoad");
+            cy.get(".enrichment-item").eq(0).should("include.text","symbiosis");
+            cy.get(".enrichment-p-value").eq(0).should("have.text","1.783148e-7");
+        })
+    })
```
heralden: This is fine, but can you also add tests which only test one specific filter, for *max p-value* and *test correction*, similar to "can filter enrichment results by ontology" checking that the count has changed (to make this obvious, you can check the count prior to setting the filter, and then after).

cypress/integration/lists_test/enrichment_test.js
heralden: Note that for "can view and download single enrichment result," you can remove the download part as the only way to download a single enrichment result is by checking it first (which you have a separate test for).

cypress/integration/lists_test/enrichment_test.js
```diff
@@ -0,0 +1,99 @@
...
+            cy.get(".list-selection").eq(0).click();
+        })
+    });
+
+    it("can view and download all enrichment results", function(){
+        cy.get(".enrichment").should("exist").within(() => {
+            cy.get('.enrichment-header').within(() => { 
+                cy.get('[type="checkbox"]').check();
+                cy.wait(500);
+                cy.get('[type="checkbox"]').should("be.checked");
```
heralden: Can you click the view/download buttons directly instead of checking all first? It shouldn't be necessary as if none are selected, it will default to all.

cypress/integration/lists_test/enrichment_test.js
```diff
@@ -0,0 +1,122 @@
...
+    it("can download all enrichment results", function(){
+        cy.get(".enrichment").should("exist").within(() => {
+            cy.get(".enrichment-category").eq(0).within(() => { 
+                cy.get(".icon-download").click();
+            })
+        })
+        cy.readFile("cypress/downloads/Enrichment Results Gene Ontology Enrichment.tsv").then(newResult => {
+            cy.readFile("cypress/fixtures/all_enrichment_result.tsv").should("eq",newResult);
+        })
+    })
```
heralden: What happened to also testing "View" here? Looks like it got lost...

cypress/integration/report_page_test.js
```diff
@@ -110,4 +114,13 @@ describe("Report Page Test", function(){
...
     })
+
+    it("can access other mines", function(){
+        cy.get(".other-mine").within(() => {
+            cy.get(".mine").should("include.text","FlyMine");
+            cy.get("img").should("exist");
+            cy.contains("Eip75B").should("exist").click();
+        })
+        cy.url().should("include","/flymine");
+    })
```
heralden: Can you also assert that the *D. melanogaster* organism is present and assert the full URL "/flymine/report/Gene/1029368" to ensure the link points to the report page and not just the mine.

heralden (approved):

PR #778, opened 2022-02-07
Linking in to ID resolver with portal.do endpoint

### LINKED ISSUE #735: Creating a new endpoint similar to the legacy portal.do
In BG we need a portal endpoint similar to that one provided by the legacy interface.
It's used by the im registry, when is acting as a splash page to import from Galaxy to InterMine.
Data are sent using the method POST.
Here a code section from the im-registry
`function mineNavButton(mine, dataToTransfer) {
  return '<form action="' + mine.url + '/portal.do" name="list" method="post">' +
    '<input type="hidden" name="externalids" value="' +
    dataToTransfer.identifiers.join(",") + '" />' +
    '<input type="hidden" name="class" value="' +
    dataToTransfer.identifierType + '" />' +
    '<input type="submit" value="Send to ' + mine.name + '" />' +
    '</form>';
}`
The endpoint should accept also the optional extraValue (organism).
The new endopoint after checking the identifiers posted, redirect to the page below
image

heralden:
Same API as the legacy portal.do documented here (sections with URLs including portal.do): https://intermine.readthedocs.io/en/latest/webapp/linking-in/#list-of-identifiers

A significant difference is that it doesn't automatically save the list (when ID resolution doesn't require manual intervention) and show the list analysis page; or show the report page when there's a single ID.

Fixes #735

### DIFF
--- a/src/clj/bluegenes/routes.clj
+++ b/src/clj/bluegenes/routes.clj
@@ -80,21 +80,21 @@
       ;; Linking in.
       ;; Handles both configured mines and the /query path.
       (wrap-params
-        (wrap-keyword-params
-          (apply compojure/routes
-                 (for [path (concat (map :namespace mines) ["query"])
-                       :let [redirect-path (str (:bluegenes-deploy-path env) "/"
-                                                (when-not (= path "query")
-                                                  path))]]
-                   (context (str "/" path) []
-                     (GET "/portal.do" {params :params}
-                        (-> (found redirect-path)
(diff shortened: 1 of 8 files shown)


heralden:
Only failing test after [3fa1eeb](https://github.com/intermine/bluegenes/pull/778/commits/3fa1eebd95d4f1215eb01bf8ed45401e5c14a219) is `1) Search Test can search with boolean search syntax` which has to do with backend/biotestmine changes.

PR #782, opened 2022-02-10
QB: Fix crash caused by adding descendent of subclass constraint

### LINKED ISSUE #773: error rerunning existing query
I was trying to rerun a query to answer
https://github.com/pombase/pombase-chado/issues/929

but I got an error

TypeError: Cannot read properties of null (reading 'toLowerCase')

    in r.
    in ul
    in li
    in r.
    in ul
    in li
    in r.
    in ul
    in div
    in Dmb
    in div
    in Kmb
    in div
    in div
    in Lmb
    in div
    in div
    in div
    in div
    in Umb
    in utb
    in main
    in div
    in BlueGenesRootErrorBoundary
    in Q.
    in vtb

### LINKED ISSUE #781: Crash caused by adding descendent of subclass constraint
where the class with the subclass constraint is not yet added to the query.

Can be reproduced on both Humanmine and Flymine by following: Select root data type Gene, expand Child Features and type constrain to CRM, expand TF Binding Sites and add Primary Identifier to view.

Reported in #773

heralden:
Also fixes the same bug when summarising a descendent.

Fixes #781
Fixes #773

### DIFF
--- a/src/cljs/bluegenes/pages/querybuilder/events.cljs
+++ b/src/cljs/bluegenes/pages/querybuilder/events.cljs
@@ -435,20 +435,24 @@
          class (im-path/class model (join "." original-path-vec))
          summary-fields (get all-summary-fields (or (keyword subclass) class))
          adjusted-views (map (partial split-and-drop-first original-path-vec) summary-fields)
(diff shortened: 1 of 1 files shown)


heralden:
Two CI runs completed without failures, but were terminated due to context deadline.

PR #845, opened 2025-06-17
Create config subdirectories for dev and prod

martinburchell:
The `.gitignore` file will ignore files like `config/dev/config.edn` and `config/prod/config.edn` but the `dev` and `prod` subdirectories don't exist in version control. This looks like an oversight.

This change ensures these directories are in git by including a `README.md` in each.
This change will also ignore Emacs backup files


martinburchell:
I don't think the CircleCI failures are down to anything I've done. I tried fixing the missing Ubuntu image but there were failures further down the line. I'll stop there.

